Issue #2, opened 2018-04-22
Request: `.flake8` configuration file support


GarisonLotus:
[Lines 112 - 114 in runway/commands/base.py](https://github.com/onicagroup/runway/blob/master/runway/commands/base.py#L112-L114) are executing a flake8 run during runway preflight specifying `exclude` directives.  Because of this, the .flake8 configuration file I have in the root of my repo is ignored. 

I've verified this at the root of a repo I'm working in by manually executing flake8, both with `--exclude=node_modules,.serverless` and without. When I include the exclude directive, my `.flake8` config file is ignored.

Can we check the filesystem for .flake8 (and maybe the other names flake8 supports) and source them if present? It seems odd we support overriding the yamllint configuration, but not flake8.

troyready:
Thanks @GarisonLotus . Added `.flake8` support in v0.13.0

PR #3, opened 2018-04-22
issue #2 potential solution

GarisonLotus:
Threw this together to get a conv started on a solution, or a conv started on why this is a stupid idea.


troyready:
Implemented in 2cf7746

PR #4, opened 2018-06-22
added quickstart cfn, docker, and docs

onica-emiller:


GarisonLotus:
You know what would also be nice to showcase here since this is a pubic repo, is an example of a CI/CD pipeline configuration that includes runway.  We already have a working one for bitbucket pipelines that we could include here. 

thoughts?

quickstarts/runway-quickstart.yml
```diff
@@ -0,0 +1,237 @@
...
+
+Conditions: 
+  DeployWindows: !Equals [!Ref OS, Windows]
+  DeployLinux: !Equals [!Ref OS, Linux]
+  DeployIamRole: !Equals [!Ref IamRole, auto/admin]
+
+
+Mappings:
+  AMIs:
+    us-east-2:
```
GarisonLotus: Only thing I would suggest would be adding other regions to your AMI map in the CFN template.

onica-emiller: @GarisonLotus agreed...see 'cfn mappings' comment a few lines above yours ^^     :)
If I can get this merged I'll add the mappings today...

PR #17, opened 2018-10-23
Improve Resilency on Approval Options when making changes to stack

blade2005:
By making it all lowercase the usable options become Y/y, v/V and anything else skips.

This addresses perhaps a small issue but one which may be confusing if you've accidentally turned on caps lock while coding.

I'm also a big believer that in cases like this there shouldn't be a difference between uppercase and lowercase options.

### DIFF
--- a/runway/embedded/stacker/providers/aws/default.py
+++ b/runway/embedded/stacker/providers/aws/default.py
@@ -155,7 +155,7 @@ def ask_for_approval(full_changeset=None, params_diff=None,
         approval_options.append('v')
 
     approve = ui.ask("Execute the above changes? [{}] ".format(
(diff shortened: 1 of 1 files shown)


troyready:
@blade2005 can you make this same PR against https://github.com/cloudtools/stacker/blob/1.5.0/stacker/providers/aws/default.py#L165

I like it, and if it's merged there or looks like it'll be merged there then I'll merge here too

blade2005:
@troyready https://github.com/cloudtools/stacker/pull/674

blade2005:
@troyready it's got merged. Will be in stacker 1.5.1 or their next release.

troyready:
Thanks! Released in v0.35.2

PR #28, opened 2019-02-27
Update quickstart.rst

elfclass32:
Fixing so this will run on Linux and Macos
Adding an extension for the first sed -i to create a backup file, and adding the -e so this will run on macos


troyready:
Thanks!

PR #33, opened 2019-03-11
improve progress messages

rgitzel:


runway/commands/env.py
```diff
@@ -336,28 +336,40 @@ def run(self, deployments=None, command='plan'):  # noqa pylint: disable=too-man
...
                         context.env_vars,
                         {'AWS_DEFAULT_REGION': context.env_region,
                          'AWS_REGION': context.env_region}
                     )
                     if deployment.get('assume-role'):
-                        pre_deploy_assume_role(deployment['assume-role'],
-                                               context)
-                    if deployment.get('account-id') or (
-                            deployment.get('account-alias')):
+                        pre_deploy_assume_role(deployment['assume-role'], context)
```
rgitzel: these next few are not changes per se, just removing newlines that aren't need anymore, and were just confusing

PR #35, opened 2019-03-12
move some commands

rgitzel:
A few commands (like `gitclean`) were set up as 'module' commands, but they aren't, they're general across the deployment.

I also moved some code around... anything that seemed specifc to a command like `test` or `gitclean` is now in the class, not the superclass.


runway/commands/modules_command.py
```diff
@@ -82,71 +79,6 @@ def determine_module_class(path, module_options):
     return load_object_from_string(module_options['class_path'])
 
 
-def get_deployment_env_vars(env_name, env_var_config=None, env_root=None):
```
rgitzel: moved to the parent

runway/commands/modules_command.py
```diff
@@ -304,24 +236,6 @@ def echo_detected_environment(env_name, env_vars):
 class ModulesCommand(RunwayCommand):
     """Env deployment class."""
 
-    def gitclean(self):
```
rgitzel: moved to the gitclean class

runway/commands/modules_command.py
```diff
@@ -401,54 +316,53 @@ def run(self, deployments=None, command='plan'):  # noqa pylint: disable=too-man
                     if deployment.get('current_dir'):
                         modules.append('.' + os.sep)
                     for module in modules:
-                        module_opts = {}
```
rgitzel: moved to a function... PyLint complained this has too many local variables

runway/commands/runway_command.py
```diff
@@ -68,54 +60,6 @@ def get_yaml_files_at_env_root(self):
         )
         return yaml_files + yml_files
 
-    def lint(self, base_dir=None, dirs_to_scan=None):
```
rgitzel: moved to the Test class

PR #36, opened 2019-03-13
replace many long filesystem and environment incantations

rgitzel:
in particular things like 
```
isinstance(self.options.get('environments',{}).get(self.context.env_name)
```
and
```
os.path.isfile(os.path.join(self.path,
                                           '.terraform-version'))
```

### DIFF
--- a/runway/module/__init__.py
+++ b/runway/module/__init__.py
@@ -6,8 +6,10 @@ import os
 import platform
 import subprocess
 import sys
+import json
+import yaml
 
-from ..util import which
+from ..util import which, better_dict_get
 
 LOGGER = logging.getLogger('runway')
 NPM_BIN = 'npm.cmd' if platform.system().lower() == 'windows' else 'npm'
@@ -98,14 +100,29 @@ class RunwayModule(object):
 
     def __init__(self, context, path, options=None):
         """Initialize base class."""
+        self.name = os.path.basename(path)
+
         self.context = context
 
+        # it would be good to remove the need for sub-classes to refer
+        #  to this directly, and have them rely on 'folder' instead
         self.path = path
+
+        self.folder = RunwayModuleFolder(path)
 
         if options is None:
             self.options = {}
         else:
             self.options = options
+
+        self.environment = better_dict_get(self.options, 'environments', {}).get(context.env_name)
+        if self.environment is not None:
+            # a boolean indicates we want (or don't want) the module, and we have no values to set
+            if isinstance(self.environment, bool):
+                if self.environment:
+                    self.environment = {}
+                else:
+                    self.environment = None
 
     def plan(self):
         """Implement dummy method (set in consuming classes)."""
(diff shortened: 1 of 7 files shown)


runway/module/__init__.py
```diff
@@ -98,15 +100,30 @@ class RunwayModule(object):
...
         """Initialize base class."""
+        self.name = os.path.basename(path)
+
         self.context = context
 
+        # it would be good to remove the need for sub-classes to refer
+        #  to this directly, and have them rely on 'folder' instead
         self.path = path
 
+        self.folder = RunwayModuleFolder(path)
```
rgitzel: defined below, hides the details of _most_ filesystem interactions

runway/module/__init__.py
```diff
@@ -98,15 +100,30 @@ class RunwayModule(object):
...
+
         if options is None:
             self.options = {}
         else:
             self.options = options
 
+        self.environment = better_dict_get(self.options, 'environments', {}).get(context.env_name)
```
rgitzel: This logic was repeated in each module... I think this is the behaviour we want for all of them.

runway/module/__init__.py
```diff
@@ -121,3 +138,53 @@ def destroy(self):
...
+    def locate_file(self, names):
+        """Given a list of files, find one that exists (if any) in the root folder."""
+        for name in names:
+            if self.isfile(name):
+                return name
+        # IDEA: it might be better to find *all* of the existing files,
+        #  and log a warning if more than one is found?
+        return None
+
+    def locate_env_file(self, names):
```
rgitzel: Some modules use 'env' some don't.... may as well let them all do it.   And we could add other folders here in the future.

runway/module/cloudformation.py
```diff
@@ -14,42 +13,6 @@
 LOGGER = logging.getLogger('runway')
 
 
-def ensure_stacker_compat_config(config_filename):
```
rgitzel: Moved these inside the class to take advantage of `folder` and to remove some params.

runway/module/serverless.py
```diff
@@ -16,33 +15,6 @@
 LOGGER = logging.getLogger('runway')
 
 
-def gen_sls_config_files(stage, region):
```
rgitzel: Moved these inside the class to take advantage of `folder` and to remove some params.

runway/module/serverless.py
```diff
@@ -103,20 +102,23 @@ def run_serverless(self, command='deploy'):
...
-                "this stage/region found (looking for one of \"%s\")",
+                "Skipping 'serverless %s' for '%s' in %s; "
+                "no configuration found for this stage/region.",
                 command,
-                os.path.basename(self.path),
-                ', '.join(gen_sls_config_files(self.context.env_name,
-                                               self.context.env_region)))
+                self.name,
+                self.context.env_region)
+            LOGGER.info("Looking for one of the following in root or 'env' folder:")
```
rgitzel: If found the list on a single line confusing visually, so broke them out on individual lines.... though lately I'm thinking this might not be valuable.

rgitzel:
Whoops!  Forgot to `lint`.... fixing....

rgitzel:
I'm going to rejig this a bit.

PR #38, opened 2019-03-14
simplify the deployments and modules menus

rgitzel:
Instead of raw JSON...
```
Configured deployments:
1: hosted_zone
2: generator, domain, sampleapp.tf


Enter number of deployment to run (or "all"): 2

Configured modules in deployment:
1: generator
2: domain
3: sampleapp.tf


Enter number of module to run (or "all"):
```

### DIFF
--- a/runway/commands/modules_command.py
+++ b/runway/commands/modules_command.py
@@ -9,7 +9,6 @@ from subprocess import check_call, check_output
 
 import copy
 import glob
-import json
 import logging
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


troyready:
I like the idea here (it's very much needed).

I think the implementation is going to need to take into account:

1) module definitions may simply be strings (where you can't call `module['path']` on them). Not sure if it's better to fix the data upfront to make it more consistent or just conditionally do the right thing (there's elsewhere in the code that we check for six.stringtypes for this same reason)
2) It's nice to show the additional module config, when it's defined, in addition to the path. Should be cleaner than the current json dump, but if you have the same module defined multiple times with different options it would be nice to be able to see them

rgitzel:
How about this?
```
Configured deployments:
1: fii, True, 6.7, fred ()
2: hosted_zone (us-east-1)
3: generator, generator, dns (us-east-1, us-west-1, us-west-2)


Enter number of deployment to run (or "all"): 3

Configured modules in deployment:
1: generator ({'foo': 'bar'})
2: generator ({'foo': 'baz'})
3: dns
```
Where the first deployment is just 
```
  - modules:
      - "fii"
      - true
      - 6.7
      - path: fred
```

troyready:
Looks great, thanks @rgitzel ! I'm torn on whether we'll want to add additional module config info into the display later, but I think you may have this correct already.

PR #40, opened 2019-03-20
allow access to runway.yml env_vars from Terraform

rgitzel:
It *seems* that you can't use these values currently?

e.g. if I have a TF file with the following
```
data "aws_route53_zone" "hosted_zone" {
  name         = "${var.app_domain}."
  private_zone = false
}
```
And then in `runway.yml` I have this in my deployment:
```
    env_vars:
      '*':
        app_domain: lab.foo.com
```
Then it doesn't work, without this code change.  

From what I can tell,  TF only has access to values in `context.env_vars` if they are prefixed by `TF_VAR_`.

### DIFF
--- a/runway/module/terraform.py
+++ b/runway/module/terraform.py
@@ -225,6 +225,9 @@ class Terraform(RunwayModule):
             for (key, val) in self.options['environments'][self.context.env_name].items():  # noqa
                 tf_cmd.extend(['-var', "%s=%s" % (key, val)])
 
+        for (key, val) in self.context.env_vars.items():  # noqa
+            tf_cmd.extend(['-var', "%s=%s" % (key, val)])
+
         if self.options.get('environments', {}).get(self.context.env_name) or (
                 workspace_tfvar_present):
             LOGGER.info("Preparing to run terraform %s on %s...",


troyready:
(reposting my internal slack response about this because I'm lazy)

The problem is the addition of OS environment variables in runway, when they generally shouldn't be used.

The thing to use here is either:
1. A per-environment tfvars file in your terraform directory, like `dev-ca-central-1.tfvars` where you define `app_domain` (this was traditionally the only way to define variable values, to keep the runway integration very light. The first part of the docs for terraform state that it's still required, though you can do it in the runway.yml now as the next sub-heading notes: https://docs.onica.com/projects/runway/en/latest/module_configuration.html#terraform), or
2. Define the variable values under a per-environment key in the `environments:` deployment or modules config (see foo being set to bar in the "Environment Values Via Runway Deployment/Module Options" sub-heading of https://docs.onica.com/projects/runway/en/latest/module_configuration.html#terraform)
The runway env_vars setting is for literal OS environment variables, which is really only relevant for other tools like kubectl where they absolutely must be set. For anything else, it really shouldn't be used
I'm sure the docs could help guide into this much better

Stuff like the env_vars setting should be a little more hidden from the general config settings. Isn't really needed unless you're doing something really different, so it shouldn't be part of the up-front information someone finds

PR #41, opened 2019-03-22
allow choosing deployment and module with CLI arguments

rgitzel:
After too many times forgetting I had to type `2` and `1` after starting Runway, I decided that this would be useful:
```
runway deploy --deployment-index 2 --module-index 1
```


rgitzel:
Need to fix this based on recent PR merges.

PR #42, opened 2019-03-25
rename a few variables for clarity

rgitzel:
"options" was confusing, as there are a number of them from different sources

### DIFF
--- a/runway/context.py
+++ b/runway/context.py
@@ -6,10 +6,9 @@ import os
 class Context(object):
     """Runway execution context."""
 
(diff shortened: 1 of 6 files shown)


runway/context.py
```diff
@@ -6,10 +6,9 @@
 class Context(object):
     """Runway execution context."""
 
-    def __init__(self, options, env_name,  # pylint: disable=too-many-arguments
```
rgitzel: Turns out this is no longer used.

PR #43, opened 2019-03-25
some refactoring

rgitzel:
two simplifications, and one case where we can fail earlier

### DIFF
--- a/runway/commands/modules_command.py
+++ b/runway/commands/modules_command.py
@@ -51,34 +51,37 @@ def assume_role(role_arn, session_name=None, duration_seconds=None,
             'AWS_SESSION_TOKEN': response['Credentials']['SessionToken']}
 
 
-def determine_module_class(path, module_options):
+def determine_module_class(path, class_path):
     """Determine type of module and return deployment module class."""
-    if not module_options.get('class_path'):
+    if not class_path:
         # First check directory name for type-indicating suffix
-        if os.path.basename(path).endswith('.sls'):
-            module_options['class_path'] = 'runway.module.serverless.Serverless'  # noqa
-        elif os.path.basename(path).endswith('.tf'):
-            module_options['class_path'] = 'runway.module.terraform.Terraform'
-        elif os.path.basename(path).endswith('.cdk'):
-            module_options['class_path'] = 'runway.module.cdk.CloudDevelopmentKit'  # noqa
-        elif os.path.basename(path).endswith('.cfn'):
-            module_options['class_path'] = 'runway.module.cloudformation.CloudFormation'  # noqa
+        basename = os.path.basename(path)
+        if basename.endswith('.sls'):
+            class_path = 'runway.module.serverless.Serverless'
+        elif basename.endswith('.tf'):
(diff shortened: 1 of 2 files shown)


runway/commands/modules_command.py
```diff
@@ -51,34 +51,37 @@ def assume_role(role_arn, session_name=None, duration_seconds=None,
             'AWS_SESSION_TOKEN': response['Credentials']['SessionToken']}
 
 
-def determine_module_class(path, module_options):
+def determine_module_class(path, class_path):
     """Determine type of module and return deployment module class."""
-    if not module_options.get('class_path'):
```
rgitzel: we don't need the value outside of this function, so we don't need to store it in the dict

runway/module/cloudformation.py
```diff
@@ -106,57 +106,56 @@ def run_stacker(self, command='diff'):  # pylint: disable=too-many-branches
...
-                        reversed(sorted(files))
-                        if command == 'destroy'
-                        else sorted(files)):
-                    if re.match(r"runway(\..*)?\.yml", name) or (
-                            name.startswith('.')):
-                        # Hidden files (e.g. .gitlab-ci.yml) or runway configs
-                        # definitely aren't stacker config files
-                        continue
-                    if os.path.splitext(name)[1] in ['.yaml', '.yml']:
-                        if not (stacker_env_file_present or self.options.get(
```
rgitzel: I can't find a reason why this would fail for a specific Stacker file, and not for all of them.  So I moved it outside the `with`

runway/commands/modules_command.py
```diff
@@ -421,13 +424,16 @@ def run(self, deployments=None, command='plan'):  # noqa pylint: disable=too-man
...
-                            getattr(
-                                determine_module_class(module_root, module_opts)(  # noqa
-                                    context=context,
-                                    path=module_root,
-                                    options=module_opts
-                                ),
-                                command)()
+                            module_class = determine_module_class(module_root,
+                                                                  module_opts['class_path'])
+                            LOGGER.info("creating instance of '%s'", module_class)
```
troyready: Let's make this LOGGER.debug, to keep the dotted class paths out of the normal logs

rgitzel: Makes sense.   Done.

runway/commands/modules_command.py
```diff
@@ -421,13 +424,16 @@ def run(self, deployments=None, command='plan'):  # noqa pylint: disable=too-man
...
                         with change_dir(module_root):
-                            getattr(
-                                determine_module_class(module_root, module_opts)(  # noqa
-                                    context=context,
-                                    path=module_root,
-                                    options=module_opts
-                                ),
-                                command)()
+                            module_class = determine_module_class(module_root,
+                                                                  module_opts['class_path'])
```
troyready: @rgitzel sorry, one more -- please change `module_opts['class_path']` to `module_opts.get('class_path')`

rgitzel: Good point. Bad habit.

Issue #44, opened 2019-03-26
Specify menu choices via the command-line


rgitzel:
**Background**
When you execute `Runway` without the `CI` environment variable then you are presented a choice of which deployments and modules should be deployed.  For example, you might see this:
```
$ runway deploy
...

Configured deployments:
1: base.cfn, databases.tf (us-east-1)
2: api.sls, frontend.sls (us-east-1, us-west-1, us-west-2)

Enter number of deployment to run (or "all"): 2

Configured modules in deployment:
1: api.sls ({...environment values...})
2: frontend.sls ({...environment values...})

Enter number of module to run (or "all"): 1
```
Those particular choices would result in _only_ the `frontend.sls` module being deployed.

This is intended to be a time-saver for developers who are working on a particular module; it saves them from waiting for Runway to go through all of the other modules to decide they haven't changed.  It's not intended to be used to deploy to "real" environments.

As nice as that is, choosing menu options can be error-prone, and it's easy to forget about, leaving you wondering why a long deploy didn't run while you went for coffee. 

**Proposal**
Allow the two menu selections to be provided on the command-line:
```
$ runway deploy  --deployment-index 2 --module-index 2
```
The values provided should mimic exactly the selections from the menus; so for instance this would result in deploying everything:
```
$ runway deploy  --deployment-index all
```
And this would deploy the second deployment entirely:
```
$ runway deploy  --deployment-index 2 --module-index all
```

PR #48, opened 2019-04-01
Specify menu choices via the command-line

rgitzel:
for https://github.com/onicagroup/runway/issues/44


rgitzel:
will do this change last

PR #50, opened 2019-04-02
refactor to clarify options and simplify repeated incantations

rgitzel:
A few things...
- encapsulate most access of the filesystem into a new `RunwayModuleFolder` class, for example to remove the many instances of `os.path.basename(path)`
- encapsulate functions related to `npm`, again to simplify things
- try to be more explicit about the various "options" (e.g. ones in the Runway file vs environments vs modules vs....)
- figure out the "environment" right away, instead of repeatedly calling `self.options.get('environments', {}).get(self.context.env_name, {})`

### DIFF
--- a/runway/module/__init__.py
+++ b/runway/module/__init__.py
@@ -6,50 +6,14 @@ import os
 import platform
 import subprocess
 import sys
+import json
+import yaml
 
-from ..util import which
+from ..util import which, better_dict_get
 
 LOGGER = logging.getLogger('runway')
 NPM_BIN = 'npm.cmd' if platform.system().lower() == 'windows' else 'npm'
 NPX_BIN = 'npx.cmd' if platform.system().lower() == 'windows' else 'npx'
-
-
-def format_npm_command_for_logging(command):
-    """Convert npm command list to string for display to user."""
-    if platform.system().lower() == 'windows':
-        if command[0] == 'npx.cmd' and command[1] == '-c':
-            return "npx.cmd -c \"%s\"" % " ".join(command[2:])
-        return " ".join(command)
-    # Strip out redundant npx quotes not needed when executing the command
-    # directly
-    return " ".join(command).replace('\'\'', '\'')
-
-
-def generate_node_command(command, command_opts, path):
-    """Return node bin command list for subprocess execution."""
-    if which(NPX_BIN):
-        # Use npx if available (npm v5.2+)
-        LOGGER.debug("Using npx to invoke %s.", command)
-        if platform.system().lower() == 'windows':
-            cmd_list = [NPX_BIN,
-                        '-c',
-                        "%s %s" % (command, ' '.join(command_opts))]
-        else:
-            # The nested app-through-npx-via-subprocess command invocation
-            # requires this redundant quoting
-            cmd_list = [NPX_BIN,
-                        '-c',
-                        "''%s %s''" % (command, ' '.join(command_opts))]
-    else:
-        LOGGER.debug('npx not found; falling back invoking %s shell script '
-                     'directly.', command)
-        cmd_list = [
-            os.path.join(path,
-                         'node_modules',
-                         '.bin',
-                         command)
-        ] + command_opts
-    return cmd_list
 
 
 def run_module_command(cmd_list, env_vars):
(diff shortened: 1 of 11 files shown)


runway/module/__init__.py
```diff
@@ -6,52 +6,16 @@
...
 
-from ..util import which
+from ..util import which, better_dict_get
 
 LOGGER = logging.getLogger('runway')
 NPM_BIN = 'npm.cmd' if platform.system().lower() == 'windows' else 'npm'
 NPX_BIN = 'npx.cmd' if platform.system().lower() == 'windows' else 'npx'
 
 
-def format_npm_command_for_logging(command):
```
rgitzel: these are now in `NpmHelper` class, in this same file

runway/module/cloudformation.py
```diff
@@ -14,42 +13,6 @@
 LOGGER = logging.getLogger('runway')
 
 
-def ensure_stacker_compat_config(config_filename):
```
rgitzel: These are inside the module class, to allow access to the `folder`

runway/module/serverless.py
```diff
@@ -2,47 +2,16 @@
...
-    RunwayModule, format_npm_command_for_logging, generate_node_command,
-    run_module_command, run_npm_install
-)
+from . import RunwayModule, run_module_command
 from ..util import change_dir, which
 
 LOGGER = logging.getLogger('runway')
 
 
-def gen_sls_config_files(stage, region):
```
rgitzel: These have been moved to inside the module class

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+        if self.environment_options is not None:
+            # it will almost always be a 'dict', but some modules (like serverless) support
+            #  a boolean, which indicates we want the module, and we have no values to set
+            if isinstance(self.environment_options, bool):
+                if self.environment_options:
+                    self.environment_options = {}
+                else:
+                    self.environment_options = None
+
+        self.module_options = better_dict_get(runway_file_options, 'options', {})
```
troyready: I can see why you're looking to simply this now. I had always made the modules grab these so there didn't need to be an explicit name.

I was trying model the module's nested `options` after Serverless Framework's `custom:` key, where any random module-type-specific stuff could go (in contrast with module options themselves, which are the common-across-module-types stuff).

Could call it `self.module_custom_options` to stick with the heritage? Or something totally different?

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
 
-    # the rest of these 'abstract' methods must have names which match
-    #  the commands defined in `cli.py`
+        # should global options be in the context instead?
```
troyready: Yes, global deployment/project options should be on the context if they end up being needed (next comment will discuss getting rid of the current need)

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+        self.project_options = {}
+        self.project_options['skip_npm_ci'] = runway_file_options.get('skip_npm_ci')
```
troyready: There were a couple of mistakes in retrospect here: having skip-npm-ci be a deployment option and not placing `skip_npm_ci` in the nested 'options' in the module's options (what I above call the `module_custom_options`).

Let's start deprecating them, by:

1. Updating the docs
2. Detecting their use, warning that they've been deprecated and will be removed in 1.0, and set the custom module options on behalf of the user for now

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+
+        self.npm = NpmHelper(self.name, self.project_options, self.context.env_vars, self.folder)
+
+        # ideally we don't need this, but there are extreme situations (like creating
+        #  a Cloudformation module inside the StaticSite module) that are currently
+        #  very hard without access to this
+        self._runway_file_options = runway_file_options
```
troyready: I'd rather go the other way and have the staticsite contort itself (recreating the env dictionary, etc), so most module don't have this around

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+
+        self.module_options = better_dict_get(runway_file_options, 'options', {})
 
-    # the rest of these 'abstract' methods must have names which match
-    #  the commands defined in `cli.py`
+        # should global options be in the context instead?
+        self.project_options = {}
+        self.project_options['skip_npm_ci'] = runway_file_options.get('skip_npm_ci')
+
+        self.npm = NpmHelper(self.name, self.project_options, self.context.env_vars, self.folder)
```
troyready: Let's not make this part of the base module class

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+    def load_json_file(self, name):
+        """Load the contents of the JSON file into a dict."""
+        with open(self.fullpath(name), 'r') as stream:
+            return json.load(stream)
+
+    def load_yaml_file(self, name):
+        """Load the contents of the YAML file into a dict."""
+        with open(self.fullpath(name), 'r') as stream:
+            # load() returns None on an existing but empty file, so provide a default
+            return yaml.load(stream) or {}
```
troyready: Make this `yaml.safe_load`

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+        """Given a list of files, find one that exists (if any) in the root or `env` folders."""
+        # first try in the root of the module folder
+        location = self.locate_file(names)
+        if not location:
+            # next try in the 'env' folder
+            env_names = [os.path.join('env', name) for name in names]
+            location = self.locate_file(env_names)
+        return location
+
+    def load_json_file(self, name):
```
troyready: Let's drop this, and change the current use of it to also use `load_yaml_file`

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+        raise NotImplementedError('You must implement the deploy() method yourself!')
 
     def destroy(self):
         """Implement dummy method (set in consuming classes)."""
-        raise NotImplementedError('You must implement the destroy() method '
-                                  'yourself!')
+        raise NotImplementedError('You must implement the destroy() method yourself!')
+
+
+class RunwayModuleFolder(object):
```
troyready: I'd like this bikeshed color to be `RunwayModulePath`

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+            return yaml.load(stream) or {}
+
+
+class NpmHelper(object):
+    """Functions to wrap around npm for use by the various modules."""
+
+    def __init__(self, module_name, project_options, env_vars, folder):
+        """Create an instance."""
+        self._module_name = module_name
+        self._project_options = project_options
```
troyready: This would become something like `module_custom_options` based on my deprecation suggestion above.

runway/module/__init__.py
```diff
@@ -60,67 +24,169 @@ def run_module_command(cmd_list, env_vars):
...
+        self._env_vars = env_vars
+        self._folder = folder
```
troyready: Bikeshed color change to `path`

runway/util.py
```diff
@@ -79,6 +79,25 @@ def merge_dicts(dict1, dict2, deep_merge=True):
...
+#
+# a YAML file can be parsed into a dict with values of None!
+#
+# for example a file containing just 'foo:' will be parsed as `{'foo': None}`
+#
+# thus we can't rely on `dict.get(key, default)` as it will return `None`
+#  instead of default in those cases :-(
+#
+# not that value-less nodes are something we should encourage, but at least
+#  using this function saves us from None errors
```
troyready: I would rather not do this for now -- can we revisit with your schematics work later?

I'd rather catch `foo: ` shenanigans as an early config/model check (ideally with a way to differentiate that from `foo: !!null`)

rgitzel: Good point. And it's only used in two places.  So for now either of them being 'empty' will trigger a crash.  Later we'll explicitly reject them as mis-configured.

Removing this function.

runway/module/__init__.py
rgitzel: I can do that by giving staticsite its own constructor.

runway/module/__init__.py
rgitzel: can probably just create it as needed, or at worst use subclass constructors

troyready:
Closing this for now due to other refactoring

PR #62, opened 2019-07-31
Use Terraform Reconfigure when initializing

cmilam87:
Remove the backend comparison and instead check if there is a failed terraform init, if so trash the folder and run init with the reconfigure option.

Always run init with the reconfigure option and Terraform will "do the right thing"


cmilam87:
I'm still working on tests to verify this works as intended, but so far it's looks good.

troyready:
Closing in favor of #63

PR #65, opened 2019-08-21
update tests to be user configurable, give config its own class

ITProKyle:
- remove default tests
- tests are now defined in the config
- config has its own class to find, load, and validate
- added a high level base class for commands (use for test command and now runway commands)
- 3 built in test types; script, cfn-lint, yamllint
- removed unused requirements
- update documentation

---

### runway.yml

```yaml
tests:
  - name: yamllint-example
    type: yamllint
    required: false
  - name: cfn-lint-example
    type: cfn-lint
    required: false
  - name: hello-world
    type: script
    args:
      commands:
        - echo "Hello World"
  - name: this-is-the-end
    type: script
    args:
      commands:
        - echo "Goodbye"
        - exit 1
```

### Output

image

### runway.yml

```yaml
---
ignore_git_branch: true
top_level_invalid: check
tests:
  - name: hello-world
    type: script
    test_invalid: check
    args:
      commands:
        - echo "Hello World"
deployments:
  - modules:
      - path: sampleapp.cfn
        mod_invalid: check
        mod_invalid2: check
    regions:
      - us-east-1
    dep_invalid: check
```

### Output

```
$ runway plan
WARNING:runway:Invalid keys found in module sampleapp.cfn have been ignored: mod_invalid, mod_invalid2
WARNING:runway:Invalid keys found in deployment deployment_1 have been ignored: dep_invalid
WARNING:runway:Invalid keys found in test hello-world have been ignored: test_invalid
WARNING:runway:Invalid keys found in runway file have been ignored: top_level_invalid
INFO:runway:Skipping environment lookup from current git branch ("ignore_git_branch" is set to true in the runway config)
...
```


src/runway/__init__.py
```diff
@@ -1,2 +1,2 @@
 """Set package version."""
-__version__ = '0.47.1'
+__version__ = '0.48.0'
```
troyready: Let's drop the version change (and associated new version section in the CHANGELOG) in favor of leaving this unreleased. I think we'll make it part of 1.0

src/runway/commands/runway/test.py
```diff
@@ -1,191 +1,50 @@
...
-from runway.embedded.stacker.awscli_yamlhelper \
-    import yaml_parse as parse_cloudformation_template  # noqa
-from runway.util import (
-    change_dir, get_embedded_lib_path, ignore_exit_code_0, use_embedded_pkgs,
-    which
-)
-
-from ..runway_command import RunwayCommand
+from ..base_command import BaseCommand
+from ...tests.regestry import TEST_HANDLERS
```
troyready: Let's rename this to `registry`

src/runway/runway.yml
```diff
@@ -0,0 +1,7 @@
+deployments:
+  - modules:
+      - path: sampleapp.cfn
+    regions:
+      - us-west-2
+    account-id:
+      dev: test
```
troyready: Can we move this file to the tests fixtures directory or templates?

ITProKyle: looks like it was one of the files I was using during dev. my bad - i can remove it completely. it has no functional significance. was just something I was using to validate parsing.

src/runway/__init__.py
ITProKyle: reverted and moved changed under unreleased in changelog

src/runway/commands/runway/test.py
ITProKyle: fixed the type for the import and the filename...

ITProKyle:
@troyready - tossing this back to you for review again post rebase, add new and missing attrs, fix for a bug I found, and inclusion of invalid keys in config log statement.

troyready:
@ITProKyle can we leave the run-python command in? I'd like to use it now for some ancillary runway examples.

(e.g., being able to have a terraform custom data source `foo.py` that can be invoked as `runway run-python foo.py` -- cross-platform, doesn't require python, etc)

ITProKyle:
> can we leave the run-python command in? I'd like to use it now for some ancillary runway examples.

@troyready - done

PR #69, opened 2019-08-28
single-binary build, npm publish, travis overhaul

ITProKyle:
- add pyinstaller spec for single binary build, depends on #65 for test command to work
- setup travis to test across multiple version of python and build on 3 platforms
- added build artifact upload to s3 for single-binary and pypi
- add automated publish to npm (installs single-binary)

Example of a full build and deploy using travis can be found here - https://travis-ci.org/ITProKyle/runway/builds/578077802

Example build is published as itprokyle-runway on npm by travis - https://www.npmjs.com/package/itprokyle-runway


docs/source/how_to_use.rst
```diff
@@ -7,9 +7,13 @@ Getting Started
   :ref:`repo-structure` )
 - Execute ``runway init`` to create a ``runway.yml`` file at the root of the
   environment (e.g. at the repo root)
+    - If runway was installed from npm, you many need to add
+      ``$projectdir/node_modules/bin`` to your ``PATH`` or supply the
+      relative path to use the runway CLI.
```
troyready: Let's drop this note -- I like the intent of it, but we'll have a few methods for install and it'll be more straightforward to just assume that the command is available on the path

docs/source/installation.rst
```diff
@@ -1,14 +1,17 @@
...
 ============
 
-Installing Runway 
+Installing Runway
 ^^^^^^^^^^^^^^^^^
 
-Runway runs on Python 2.7 and Python 3.5+.
+Runway runs on Python 2.7 and Python 3.5+. It is also packaged as a
+single-binary bundled with python for distribution through non-python
+package managers.
```
troyready: (not a blocker for this PR) prior to the 1.0 release, we should update this to be biased to curl/wget/iwr (more like "just download and run it") with more "advanced" deployments available via pip/npm.

npm/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "undefined",
```
troyready: Just cosmetic I think, but name this ^ "runway" I think

setup.py
```diff
@@ -34,7 +34,7 @@
     # https://github.com/virtuald/pyhcl/pull/57
     'pyhcl~=0.3',
     'six',
-    'typing',
+    'typing;python_version<"3.5"',
```
troyready: Should be `<"3.6"`, right?

npm/package.json
ITProKyle: fyi, I set it to undefined to default away from runway. The name is set based on a travis env var when preparing to publish allowing forks to publish a clone from their fork as needed. but, we can default it to runway.

https://github.com/ITProKyle/runway/blob/feature/single-binary/Makefile#L57

setup.py
ITProKyle: > NOTE: in Python 3.5 and later, the typing module lives in the stdlib, and installing this package has NO EFFECT. To get a newer version of the typing module in Python 3.5 or later, you have to upgrade to a newer Python (bugfix) version.

https://pypi.org/project/typing/

troyready: Ok, you're right, I think it was one version later -- I think the Pipfile ^ can be updated then to match.

docs/source/installation.rst
ITProKyle: Unless we have the curl/wget/iwr endpoint, I'm inclined to leave it as-is and we can rewrite the block with the appropriate instructions when 1.0 is released

troyready: Actually - @ITProKyle please do drop this doc change and we'll stage it in a separate branch/PR (need to time its merging to master AKA-docs-latest with binary publishing

setup.py
ITProKyle: updated Pipfile to match

npm/package.json
ITProKyle: defaulted name to `runway` in package.json

docs/source/installation.rst
ITProKyle: reverted changes to installation.rst

docs/source/how_to_use.rst
ITProKyle: removed note

Issue #67, opened 2019-08-27
Improve error message when shelling out to commands that don't exist


troyready:
When [commands are executed](https://github.com/onicagroup/runway/blob/v0.47.1/runway/util.py#L205) (e.g. `yarn build` specified as the build_steps of a staticsite), a stacktrace is generated if the command isn't found.

This is expressed as a subprocess `FileNotFoundError` exception, which doesn't point the user to the real problem ("You told me to run `yarn build` but you don't have `yarn` installed").

We should catch the exception and print a more helpful error message (noting the failed command) before exiting 1.

troyready:
Fixed in v0.47.2

PR #70, opened 2019-08-29
don't re-check for commands when fixing windows commmands

cmilam87:
when calling `fix_windows_command_list`, if there is both a .cmd and non-extension executable, (e.g. `yarn.cmd` and a file `yarn` that is also executable `os.access(fpath, os.X_OK)`) then the if statement will return false and the wrong command will be used on windows when using subprocess. In this case `yarn` will be used instead of the proper `yarn.cmd` and subprocess will error with `FileNotFoundError: [WinError 2] The system cannot find the file specified`


cmilam87:
Added in a fix for issue #67 as well.

PR #76, opened 2019-09-10
Add package support for Serverless

cmilam87:
WIP
Add option to create a build package once and have it deployed to multiple environments.

### DIFF
--- a/integration_tests/runner.py
+++ b/integration_tests/runner.py
@@ -2,6 +2,7 @@
 from __future__ import print_function
 import os
 import logging
+import argparse
 from integration_test import IntegrationTest
 from util import (execute_tests, import_tests)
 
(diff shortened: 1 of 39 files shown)


troyready:
@cmilam87 the basic sls test execution is working now with my latest patch. I'm not 100% sure what part of it fixed the directory error, but I believe it was some combination of:
* a bug in the handler path parsing (see `extend`->`append`), possibly causing an early exception raising
* the environment variables not being updated consistently between command invocations
* some unknown interaction with the classes; I cleaned up every occurrence of `self` that didn't seem to refer to the current class. I _think_ it's a little more clear now (though I'm not married to any of the new parameter names/ordering
* something totally different?

Basic next steps of which I'm aware:
* My changes to the classes to fix the sls issues will have broken the terraform ones completely; they'll need some light updates to match the new syntax
* The tests need verification that the expected action is happening (i.e. that the same zip was used in test as was created in dev)

cmilam87:
I like and appreciate the changes you've made. It makes things a lot more clear.

I'll work on fixing the Terraform tests and somehow verifying that the correct zip is used between dev/test.

integration_tests/runner.py
```diff
@@ -2,39 +2,41 @@
 from __future__ import print_function
 import os
 import logging
+import argparse
```
troyready: I think this import can be dropped

cmilam87:
@troyready in order to make Travis green - I've got to refactor some of the staticsite stuff - I didn't know if we wanted to do that now or not.

troyready:
@cmilam87 do you have a better suggestion to fix it?

cmilam87:
> @cmilam87 do you have a better suggestion to fix it?

Nope. 😄

PR #84, opened 2019-10-24
drop current_dir option; bias top/deployment options towards hyphens

troyready:
Also threw in better error handling for git branch detection when in detached-head mode

### DIFF
--- a/docs/source/module_configuration.rst
+++ b/docs/source/module_configuration.rst
@@ -158,7 +158,7 @@ and/or for a group of modules:
       - modules:
           - path: mytfmodule
           - path: anothermytfmodule
-        module-options:  # shared between all modules in deployment
+        module_options:  # shared between all modules in deployment
           terraform_version:
             "*": 0.11.13  # applies to all environments
             # prod: 0.9.0  # can also be specified for a specific environment
@@ -241,7 +241,7 @@ and/or for a group of modules:
       - modules:
           - path: mytfmodule
           - path: anothermytfmodule
-        module-options:  # shared between all modules in deployment
+        module_options:  # shared between all modules in deployment
           terraform_backend_config:
             bucket: mybucket
             region: us-east-1
@@ -277,7 +277,7 @@ and/or for a group of modules:
       - modules:
           - path: mytfmodule
           - path: anothermytfmodule
-        module-options:  # shared between all modules in deployment
+        module_options:  # shared between all modules in deployment
           terraform_backend_config:
             region: us-east-1
           terraform_backend_cfn_outputs:
@@ -311,7 +311,7 @@ and/or for a group of modules:
       - modules:
           - path: mytfmodule
           - path: anothermytfmodule
-        module-options:  # shared between all modules in deployment
+        module_options:  # shared between all modules in deployment
           terraform_backend_config:
             region: us-east-1
           terraform_backend_ssm_params:
@@ -664,7 +664,7 @@ and/or for a group of modules:
       - modules:
           - path: myk8smodule
           - path: anotherk8smodule
-        module-options:  # shared between all modules in deployment
+        module_options:  # shared between all modules in deployment
           kubectl_version:
             "*": 1.14.5  # applies to all environments
             # prod: 1.13.0  # can also be specified for a specific environment
@@ -687,7 +687,7 @@ or an absolute one. E.g.::
               kubectl_version:
       - regions:
           - us-east-1
-    env-vars:
+    env_vars:
       staging:
         KUBECONFIG:
           - .kube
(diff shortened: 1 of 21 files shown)


src/runway/commands/modules_command.py
```diff
@@ -354,7 +352,7 @@ def run(self, deployments=None, command='plan'):  # noqa pylint: disable=too-man
         if deployments is None:
             deployments = self.runway_config['deployments']
         context = Context(env_name=get_env(self.env_root,
-                                           self.runway_config.get('ignore_git_branch', False)),
+                                           self.runway_config.get('ignore-git-branch', False)),
```
ITProKyle: With the `Config` class, all possible options have a value so we don't have to use `.get` here and instead use the attribute directly.

```suggestion
                                           self.runway_config.ignore_git_branch),
```

Also, since this is `.get`ing a class attribute, it does not have hyphens and will need to be reverted to underscores.

src/runway/commands/runway/whichenv.py
```diff
@@ -31,5 +31,5 @@ def execute(self):
         logging.getLogger('runway').setLevel(logging.ERROR)  # suppress warnings
         print(get_env(
             self.env_root,
-            self.runway_config.get('ignore_git_branch', False)
+            self.runway_config.get('ignore-git-branch', False)
```
ITProKyle: With the `Config` class, all possible options have a value so we don't have to use `.get` here and instead use the attribute directly.

```suggestion
            self.runway_config.ignore_git_branch),
```

Also, since this is `.get`ing a class attribute, it does not have hyphens and will need to be reverted to underscores.

src/runway/config.py
```diff
@@ -351,18 +335,30 @@ def __init__(self, deployment):
...
+            if not deployment.get('modules'):
+                LOGGER.error('No modules have been defined in your Runway '
+                             'deployment.')
+                sys.exit(1)
+            modules = deployment.pop('modules')
         self.modules = ModuleDefinition.from_list(
-            deployment.pop('modules', [])  # can be none if current_dir
+            modules
         )  # type: List[ModuleDefinition]
+        self.current_dir = deployment.pop('current_dir', False)  # type: bool
```
ITProKyle: This will always be `False` since its being `.pop`ed on line 344. I don't think we need this anymore though. It was only used in `module_commands` but those sections have been removed.

src/runway/config.py
```diff
@@ -281,19 +280,6 @@ class DeploymentDefinition(ConfigComponent):  # pylint: disable=too-many-instanc
...
-    Example:
-      .. code-block:: yaml
-
-        deployments:
-          - current_dir: true
-            regions:
-              - us-west-2
-            assume-role:
-              arn: arn:aws:iam::0000:role/role-name
-
```
ITProKyle: We should describe the new `modules: ['./']` design pattern in place of this section. It may not be as intuitive to some people as `current_dir` was.

src/runway/config.py
```diff
@@ -536,7 +532,8 @@ def load_from_file(cls, config_path):
             config_file = yaml.safe_load(data_file)
             result = Config(config_file.pop('deployments'),
                             config_file.pop('tests', []),
-                            config_file.pop('ignore_git_branch', False))
+                            (config_file.pop('ignore-git-branch', False) or
+                             config_file.pop('ignore_git_branch', False)))
```
ITProKyle: Should we use a consistent method for handing `-/_`?

```suggestion
                            config_file.pop('ignore_git_branch',
                                            config_file.pop('ignore-git-branch',
                                                            False))
```

docs/source/module_configuration.rst
```diff
@@ -158,7 +158,7 @@ and/or for a group of modules:
       - modules:
           - path: mytfmodule
           - path: anothermytfmodule
-        module_options:  # shared between all modules in deployment
+        module-options:  # shared between all modules in deployment
           terraform_version:
```
ITProKyle: Should these options support `-/_` as well? It _could_ get confusing knowing which to use if they both work in some places but not all.

troyready: @ITProKyle As it stands from this PR, there will be some consistency in:
* All top-level & deployment options are documented with hyphens, but support being used with either hyphens or underscores.
* All module options only support underscores (and haven't been documented or used with hyphens in the past)

I'm hesitant to make a practice/requirement of having the various module options support dual-format options as well, but don't love the current situation either.

Maybe it's better to rip off the bandaid and change the top level ones (`account-id`->`account_id`, etc), autocorrecting hyphens to underscores for now (with a deprecation warning about future removal). Thoughts?

ITProKyle: We could also bias everything toward underscores instead of hyphens so everything is consistent in the docs and leave top level the only thing that can also handle hyphens. We can leave this out of the docs as a "hidden feature" since its YMMV. This may suffice to mitigate confusion of what can/can't do it without having to drop support for things that use hyphens right now.

Essentially your idea of "rip off the bandaid" but not deprecating support for hyphens at the top level. At least not at this point. It can always be revisited in a few months.

PR #101, opened 2019-11-26
update error message for test execution with no defined tests

troyready:
Avoids semantic issue about deprecated message and gives more direction on how to fix the error

### DIFF
--- a/src/runway/commands/runway/test.py
+++ b/src/runway/commands/runway/test.py
@@ -9,6 +9,7 @@ References:
     - :ref:`Defining Tests<defining-tests>`
 
 """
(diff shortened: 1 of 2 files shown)


blade2005:
Looks good to me @troyready. I'm happy with those changes and you even gave a bit of help on making tests.

src/runway/commands/runway/test.py
```diff
@@ -29,7 +30,18 @@ def execute(self):
         if not test_definitions:
             LOGGER.warning('Use of "runway test" without defining '
```
blade2005: Should this still be warning or should it be error?

PR #102, opened 2019-11-26
use bundled yamllint when running tests in single binary

troyready:
Since yamllint is getting pulled in anyway, this will allow the bundled version to be used in single-binary mode

### DIFF
--- a/src/runway/tests/handlers/yaml_lint.py
+++ b/src/runway/tests/handlers/yaml_lint.py
@@ -3,6 +3,8 @@
 import glob
 import logging
 import os
+import sys
+import tempfile
 from typing import Dict, Any, List  # pylint: disable=unused-import
 
 from runway.tests.handlers.base import TestHandler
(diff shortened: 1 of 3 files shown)


src/runway/tests/handlers/yaml_lint.py
```diff
@@ -49,13 +67,37 @@ def handle(cls, name, args):
...
+        else:
+            # traditional python execution
+            yl_cmd = "yamllint " + ' '.join(yamllint_options)
         with change_dir(base_dir):
             ScriptHandler().handle(
                 'yamllint',
                 {'commands': [yl_cmd]}
             )
+            if getattr(sys, 'frozen', False):
+                os.remove(temp_path)
```
ITProKyle: This should probably be wrapped in `try: finally` to ensure the temp file is always removed.

```suggestion
        with change_dir(base_dir):
            try:
                ScriptHandler().handle(
                    'yamllint',
                    {'commands': [yl_cmd]}
                )
            finally:
                if getattr(sys, 'frozen', False):
                    os.remove(temp_path)
```

But, would have to double check the functionality to ensure that exceptions raised in `ScriptHandler`, intended to halt execution, still work as intended.

src/runway/tests/handlers/yaml_lint.py
```diff
@@ -28,13 +30,29 @@ def get_yaml_files_at_path(provided_path):
         )
         return yaml_files + yml_files
 
+    def get_yamllint_options(self, path, quote_paths=True):
```
ITProKyle: should this be a `@classmethod` or even a `@staticmethod`? since it doesn't care about the state of the instance it wouldn't need `self`. it also doesn't care about the state of the class so it wouldn't need `cls` other than to call other static methods.

ITProKyle:
It may be worth noting why this is being done this way in a comment somewhere to add context.

src/runway/tests/handlers/yaml_lint.py
troyready: I did some basically testing with a finally block and it looks to work as desired with the sys.exits

PR #103, opened 2019-12-02
Bugfix - Multiple CDK stacks

voodooGQ:
Previously if you had multiple CDK stacks in a deployment you would be presented with an error asking you to choose which stack you wanted deployed. With a default wildcard of '*' on both the deploy and destroy commands, we can prevent this error and ensure all stacks are being deployed.


troyready:
@voodooGQ see the travis error for a linting fix (it wants `if 'foo' in ['foo', 'bar']` instead of 2 comparisons)

Also, I'm wondering if we should put the asterisk at the end of the options (not sure if its position matters).

Finally, this is probably a good one to write an integration test on, so we catch any cdk command syntax changes that happen

voodooGQ:
@troyready sorry just seeing this. 

In my testing the position of the asterisk didn't end up mattering, I put it is close to the primary command as possible just for easy groking. Happy to move it to the end though if you'd prefer.

Still getting up-to-speed with Python linting/testing but I agree, this would be a good test candidate. I'll see what I can spin up.

voodooGQ:
@troyready Addressed as requested. Integration test added that takes a CDK module with 2 stacks and verifies deployment. Verified via `make install`

PR #108, opened 2019-12-11
Parallel Execution Integration Tests

voodooGQ:
Adds two integration tests to verify functionality when running parallel region deployments. The first test verifies that it runs successfully with the `CI` variable set (which is a requirement of Runway), and that it successfully runs in series when the `CI` variable is not set.


voodooGQ:
@ITProKyle going to summon you for review on this one as well :)

PR #109, opened 2019-12-12
Adding basic source processor for the runway path.

voodooGQ:
Adds the capability of remote resources (git) as path items in the Runway configuration file. Abstracts out the concept of a Path object to be extended upon in future releases (such as S3).

Full unit test suite as well as baseline integration test to verify directory caching.


src/runway/commands/modules_command.py
```diff
@@ -322,6 +331,65 @@ def validate_account_credentials(deployment, context):
         validate_account_alias(boto3.client('iam', **boto_args),
                                account_alias)
 
+class SourceProcessor(object):
```
ITProKyle: should we treat this as a `Path` class rather than just a source processor? We could do more here like moving the logic that exists later to ensure the path is a dir and determining the module type since it partially pertains to the provided path.

```
class Path(object):
	"""Metadata and methods for a module path."""

	def __init__(self, path, cache_dir=None):
        # type: (str, Optional[str]) -> None:
        """Initialize class.

        Args:
            path: The path to a module provided in the runway config.
            cache_dir: Optionally supply a preferred cache directory.

        """
        self._path = path  # private, unresolved path
        self.cache_dir = cache_dir or '.runway'

    @property
    def module_type(self):
        """Logic to determine the module type from the path provided."""

    @property
    def source_type(self):
        """Local, git, etc."""

    @property
    def uri(self):
        """Parsing for git uri."""

    def resolve(self):
        """Returns the resolved path and fetches if needed."""
```

Going with this approach, it would also be logical to split to its own file. I would also consider consuming this in `config.ModuleDefinition` if we wait to `resolve()` here.

Thoughts?

src/runway/commands/modules_command.py
```diff
@@ -322,6 +331,65 @@ def validate_account_credentials(deployment, context):
...
+    def __init__(self, path, cache_dir=None):
+        if not cache_dir:
+            cache_dir = os.path.expanduser("~/.runway_cache")
+
+        self.cache_dir = cache_dir
+        self.path = path
+        self.config = self.get_config(self.path)
+        self.create_cache_directory()
+        self.get_source()
```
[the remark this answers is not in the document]
voodooGQ: I don't have any strong opinions one way or another. Speeding up runtime is  always advantageous, especially when we're dealing with all these infra spinups. At the moment my focus is getting it up and working from a base level and then refactoring down from there. @troyready Do you have any strong opinions?

src/runway/commands/modules_command.py
voodooGQ: I like that idea, I especially like the idea of getting this moved to it's own separate file here eventually. These classes are getting quite large and could get unruly if we keep adding.

I'll keep plugging away until I have a working model and then we'll start working in the refactors.

src/runway/path.py
```diff
@@ -0,0 +1,119 @@
...
+
+        source = split_source_location[0]
+        temp_location = split_source_location[1]
+
+        uri, location = self.__parse_uri_and_location(temp_location)
+        location, options = self.__parse_location_and_options(location)
+
+        return [source, uri, location, options]
+
+    def configuration(self):
```
ITProKyle: This may be better served as an `@property`.

src/runway/sources/git.py
```diff
@@ -0,0 +1,100 @@
...
+LOGGER = logging.getLogger('runway')
+
+
+class Git(Source):
+    """ Git type Path Source """
+
+    def __init__(self, config, cache_dir=None):
+        LOGGER.info(config)
+        self.config = config
+        self.cache_dir = cache_dir
```
ITProKyle: This looks like something that could belong in the parent class.

src/runway/path.py
```diff
@@ -0,0 +1,119 @@
...
+        self.module_root = self.__get_module_root_dir(module)
+
+    def parse(self, module):
+        """ Retrieve the source and location of the path variable. """
+        source = 'local'
+        uri = ''
+        location = ''
+        options = ''
+
+        split_source_location = module.get('path').split('::')
```
ITProKyle: As is, it will thrown an error if `module['path']` does not exist because `.get()` will return `None` which does not have the method `split()` so either do this

```suggestion
        split_source_location = module.get('path', '').split('::')
```

or this with handling for `KeyError` if path does not exist. Which should result in `sys.exit(1)` after an error message is logged.

```suggestion
        split_source_location = module.get['path'].split('::')
```

src/runway/path.py
```diff
@@ -0,0 +1,119 @@
...
+            location = split_source_location[0]
+            return [source, uri, location, options]
+
+        source = split_source_location[0]
+        temp_location = split_source_location[1]
+
+        uri, location = self.__parse_uri_and_location(temp_location)
+        location, options = self.__parse_location_and_options(location)
+
+        return [source, uri, location, options]
```
ITProKyle: When returning multiple values, it should be a tuple (immutable) instead of a list (mutable).

This can be done with this

```suggestion
        return source, uri, location, options
```

which is like returning this but you don't need the parenthesis

```python
        return (source, uri, location, options)
```

src/runway/commands/modules_command.py
```diff
@@ -322,6 +331,65 @@ def validate_account_credentials(deployment, context):
...
+    def get_config(self, path):
+       conf = {}
+       params = path.split('>')
+       conf['uri'] = params[0]
+
+       for x in range(1, len(params)):
+           key, value = params[x].split('=')
+           conf[key] = value
+
+       return conf
```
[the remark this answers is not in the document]
voodooGQ: So it tried to do this in the most current version, however, pylint yelled at me and told me it was "An unnecessary comprehension". I put it back to a for loop and it stopped with the error. Maybe with the way I'm handling it now (lot of refactors) it's no longer relevant?

src/runway/commands/modules_command.py
voodooGQ: Latest refactors have a lot more abstractions. Take a peek and let me know if there's other things we could be doing better.

src/runway/path.py
```diff
@@ -0,0 +1,119 @@
...
+        """ Transform object into configuration settings for remote sources. """
+        conf = {
+            'source': self.source,
+            'location': self.location,
+            'uri': self.uri,
+            'options': self.options
+        }
+        return conf
+
+    def __parse_uri_and_location(self, uri_loc_str):
```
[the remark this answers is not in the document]
voodooGQ: Can you go into the name collision piece a bit more? Based on my previous testing experience we shouldn't ever call private methods, just BDD the outcome of the public methods. Exposing more of the api through a public interface can confuse the intention of the Class/Instance. e.g. If I don't anticipate it being called by outside resources it should stay private.

If that is an antipattern in Python due to name collisions though I will have to change my tact, I just want to understand it better.

src/runway/path.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if len(split_uri_location) == 3:
+            location_string = split_uri_location[2]
+
+        return [
+            '//'.join([split_uri_location[0], split_uri_location[1]]),
+            location_string
+        ]
+
+    def __parse_location_and_options(self, loc_opt_str):
```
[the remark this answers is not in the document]
voodooGQ: TIL about @classmethod, thank you!

src/runway/sources/source.py
```diff
@@ -0,0 +1,17 @@
+"""
+    Abstract parent class for a 'Source' type object.
+    Allows us to specify specific remote sourced resources for out application
+    (Git, S3, ect.)
+"""
+class Source(object):
```
ITProKyle: I would add a `fetch()` (or whatever we want to name it) method here to be implemented in subclasses as the single method used across all of them for getting.

```python
    def fetch(self):
        """Retrieve remote source.

        To be implemented in each subclass.

        """
        raise NotImplementedError
```

src/runway/path.py
```diff
@@ -0,0 +1,119 @@
...
+        split_location_options = loc_opt_str.split('?')
+        location = split_location_options[0]
+        options = {}
+
+        if len(split_location_options) == 2:
+            options = self.__parse_options_dict(split_location_options[1])
+
+        return [location, options]
+
+    def __parse_options_dict(self, options_str):
```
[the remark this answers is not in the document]
voodooGQ: TIL about `@staticmethod` :)

src/runway/path.py
voodooGQ: So this one is a little weird. My pylint was giving up issues because I didn't have enough public methods (2) on the class. I converted this to appease that... but it seems like a really weird requirement. In the past I've used a single public method for a lot of Service objects in other languages such as `call()`. 

Is that a pylint requirement we should look to correct?

voodooGQ: And TIL about `@property` :D

src/runway/sources/git.py
voodooGQ: Good call, will move it up

src/runway/path.py
voodooGQ: What would be the preferred approach? The first example is very similar to what I've done with past languages, so I'd lean towards that.

src/runway/sources/source.py
voodooGQ: This is very similar to how I used to do it in Ruby, good call.

src/runway/path.py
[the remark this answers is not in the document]
ITProKyle: THB, you probably have a better understanding of what should be tested/shouldn't be tested than I do.

I used unittest more as an example.

But, python (the interpreter) does not have a true private/protected method. Using underscores in front of names is only a convention that is only "enforced" by linters.

I guess using `__` isn't _wrong_. It just functions differently.
[More detailed info about the differences](https://dbader.org/blog/meaning-of-underscores-in-python)

src/runway/path.py
ITProKyle: It looks like what you have _should_ be able to handle the fist option as is so go for it.

src/runway/path.py
ITProKyle: you can always add `#  pylint: disable=too-few-public-methods` to the same line as the class definition if the class is intended to only have 1.

src/runway/commands/modules_command.py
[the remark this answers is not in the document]
voodooGQ: bump @troyready

src/runway/path.py
```diff
@@ -0,0 +1,130 @@
...
+
+LOGGER = logging.getLogger('runway')
+
+
+class Path(object):  # pylint: disable=too-many-instance-attributes
+    """Runway configuration 'path' settings object."""
+
+    def __init__(self, module, env_root, cache_dir=None, git_source_class=Git):
+        # type: (str, str, Optional[str])-> Path
+        """Initialize."""
```
ITProKyle: add documentation for `Args` (minus `self` or `cls`) - in other functions/methods as well.

src/runway/path.py
```diff
@@ -0,0 +1,130 @@
+"""Runway configuration 'path' settings."""
```
ITProKyle: Can we add documentation here to be pulled in by `automodule`? It can then be added to `docs/source/runway_config.rst` under the `Module` header unless you can think of a better place to put it.

May also be worth adding a link in the docs on `runway.config.ModuleDefinition` where it talks about `path`.

The docs here should touch on local paths (`./` and `sampleapp.cfn`) - can likely move any in-depth path details from `runway.config.ModuleDefinition` to here. it should also mention _remote sources_ and link to them. There should be examples of the different ways to define `path`

src/runway/sources/git.py
```diff
@@ -0,0 +1,114 @@
+"""'Git' type Path Source."""
```
ITProKyle: Can we add documentation here to be pulled in by `automodule`? It can then be added to `docs/source/runway_config.rst` under the `Module.Path` header unless you can think of a better place to put it.

It should describe the different components of the syntax for defining a git source and show examples.

voodooGQ:
@ITProKyle I added documentation as you requested based on what we discussed this morning. Feel free to review and let me know if there's any changes you'd like to be made.

docs/source/runway_config.rst
```diff
@@ -30,6 +30,45 @@ Module
...
+``git://github.com/foo/bar.git``: The protocol and URI address of the git
+repository
+
+``//`` **(optional)**: Logical separator of the URI from the remaining path
+string
+
+``my/path`` **(optional)**: The relative path from the root of the repo
+where the module is housed
+
+``?`` (optional): Logical separator of the path from the options
```
ITProKyle: Why are the other `(optional)`s bold but not this one?

src/runway/path.py
```diff
@@ -0,0 +1,273 @@
...
+                - path: git::git://github.com/your_handle/your_repo.git//my-module.cfn
+
+
+    The ``path`` structure is based on the
+    encoding found in
+    `Terraform modules <https://www.terraform.io/docs/modules/sources.html>`_.
+
+    The values parsed from the string are as follows:
+
+    **source**: Determine if the source is local or remote. The initial
```
ITProKyle: Would this be better as a header or list item?


`- **source**: Det...`

or you can use `rubric` to create a "header" that does not appear in the toc.

```
    .. rubric:: Source

    Det...
```

src/runway/path.py
```diff
@@ -0,0 +1,273 @@
...
+    prefix is used to determine this separated by `::` in the string.
+    A path is considered local if it contains no source type value.
+
+    Example:
```
ITProKyle: If you make this a rubric, it will stand out more. or, does it stand out too much like that?

```
.. rubric:: Example
.. code-block:: yaml
```

docs/source/runway_config.rst
```diff
@@ -30,6 +30,45 @@ Module
 
 .. autoclass:: runway.config.ModuleDefinition
 
+Path
+^^^^
+
+.. autoclass:: runway.path.Path
```
ITProKyle: Since `Path` is not being defined as a mapping in the config file, I don't see an advantage to including the kwargs on this doc page. If you use `.. automodule:: runway.path.Path` it will exclude the `__init__` docstring with the kwargs.

src/runway/commands/modules_command.py
```diff
@@ -525,14 +520,11 @@ def _deploy_module(self, module, deployment, context):
...
         if deployment.get('module_options'):
             module_opts['options'] = deployment['module_options'].copy()  # noqa
-        if isinstance(module, six.string_types):
-            module = {'path': module}
-        if path_is_current_dir(module['path']):
-            module_root = self.env_root
-        else:
-            module_root = os.path.join(self.env_root, module['path'])
+
+        path = Path(module, self.env_root, os.path.join(self.env_root, '.runway_cache'))
```
ITProKyle: Would it be better to initialize this class when the config file is being parsed?

src/runway/path.py
voodooGQ: I tried to do the list item style initially but the formatting gets really strange and all over the place. I tried the rubric header as you suggested and think th is will be a good alternative. Updating.

src/runway/path.py
voodooGQ: It actually appears to already format that way without specifying the rubric. 
image

docs/source/runway_config.rst
voodooGQ: Agreed, updating.

src/runway/commands/modules_command.py
voodooGQ: I didn't want to mess around too much in modules_command initially just because I'm so new to the ecosystem and this is already a fairly large PR. I think this is worth looking at as a refactor effort where I could garner a little more guidance from you around what's advantageous or not.

PR #116, opened 2019-12-24
runway config file variables/lookups

ITProKyle:
Support for runway to resolve variables in the runway config file using lookups to simplify long/repeated config files where yaml anchors just arn't enough. Also removes the need to use maps when defining values for different environments.

# What Changed

## Added
- variable resolution in the runway config file
    - uses lookups to resolve value
        - resolves from environment variables, variables file, or variables definition
        - lookups can take arguments in addition to a query
            - parsing is part of the base class to standardize syntax
            - used to provide default values if the query fails and/or transform the data
    - only resolvable in specific areas of the config file (see docs for details)
    - some environment variables can only be used during processing of a module since they are set during processing
- `variables` top-level directive to the runway config file
    - explicitly define the path/name to a variables file instead of using the default path/names
    - define variables directly in the runway config file
        - if this is used with a variables file, what is defined in the runway config takes precedence

## Changed
- `.cfn` modules no longer require `deployments[].environments.$DEPLOY_ENVIRONMENT` to be deployed when opting to not use a `$DEPLOY_ENVIRONMENT-$AWS_REGION.env` file if variables/lookups are used


troyready:
@ITProKyle can you break the pyhcl fix out into a separate PR? I'd like to merge that in now while we finalize this one.

troyready:
I like the variable capability a lot here. The only part that isn't sitting right with me is the interaction with the variables & the current "environments:".

e.g., for the new doc environments example:
```
        environments:
          region: ${env:AWS_REGION}
          image_id: ${var:image_id.${env:DEPLOY_ENVIRONMENT}}
```
region & image_id end up being used as "parameters" (tf_var_, stacker -e, etc) for their modules, which doesn't really match the "environment" name.

I'm wondering if it makes sense to make a new parameters (not in love with any particular name) config key that could be used directly (like ^), and that the "environments:" would coalesce into? Or something in that direction?

ITProKyle:
> @ITProKyle can you break the pyhcl fix out into a separate PR? I'd like to merge that in now while we finalize this one.

opened #123 to address this

ITProKyle:
> make a new parameters (not in love with any particular name) config key that could be used directly (like ^), and that the "environments:" would coalesce into?

@troyready That makes sense. Would we want to sunset the use of `environments` (with a warning that it will continue to work until v2) in that case? If that is what we want to plan on doing, what would the interim precedence be? `parameters > environments`?

**parameters** or **arguments** could be viable names for the new config key. open to other considerations as well.

src/runway/commands/modules_command.py
```diff
@@ -520,26 +522,45 @@ def _process_modules(self, deployment, context):
...
         module_opts = load_module_opts_from_file(module_root, module_opts)
 
         LOGGER.info("")
         LOGGER.info("---- Processing module '%s' for '%s' in %s --------------",
-                    module['path'],
+                    module.path,
                     context.env_name,
                     context.env_region)
         LOGGER.info("Module options: %s", module_opts)
+        module_opts['_using_vars'] = bool(self.runway_vars)
```
troyready: @ITProKyle as part of our config discussion: I'd like to refactor so this isn't needed (i.e. the stacker module just uses the "parameter" values provided to it)

ITProKyle:
if #134 is added first, we can drop some of the stuff from there (like `variables.py`) and just reusing what is in `runway.cfngin` I avoided using it when it was `embedded/stacker` so core functionality wasn't relying on something embedded but now that its permanent, I would feel comfortable taking advantage of it.

ITProKyle:
@troyready - this should be ready for re-review. The CFNgin change is now incorporated.

PR #121, opened 2020-01-03
kbenv integration test

edgarpoce:
**Goal**
Add an integration test for kbenv command.

**Implementation**
The test creates an EKS cluster with terraform and checks whether the default namespace is Active with kbenv. On teardown the stack is destroyed.
Added an optional environment variable to to filter the integration tests. This simplifies development because it allows to run a single command test optionally.


troyready:
@edgarpoce how would you feel about using the the gen-sample template here instead? I don't want to overcomplicate the test setup, but also not having to "maintain" the k8s code in multiple places could be nice.

integration_tests/test_commands/tests/test_kbenv.py
```diff
@@ -0,0 +1,76 @@
...
+        # deploy stack
+        check_output(
+            ['runway', 'deploy', '--tag', 'eks'],
+            cwd=self.stack_path,
+            env=self.env
+        ).decode()
+
+        # create kubeconfig
+        response = check_output(
+            ['aws', 'eks', 'update-kubeconfig', '--name', 'k8s-dev',
```
troyready: Can we swap `'aws'` here with `'runway', 'run-aws', '--'` to increase the incidental runway testing here?

integration_tests/test_commands/tests/test_kbenv.py
```diff
@@ -0,0 +1,76 @@
...
+        self.tmp_stack = tempfile.TemporaryDirectory(dir=tmp_root)
+        self.stack_path = self.tmp_stack.name + '/k8s-tf-infrastructure'
+        self.env = dict(
+            os.environ,
+            CI='1',
+            DEPLOY_ENVIRONMENT='dev',
+            KUBECONFIG='{0}/.kubeconfig'.format(self.stack_path))
+
+        # create project from k8s-tf-repo sample
+        check_output(
```
troyready: It *might* be worth considering swapping these `check_output(...).decode()` calls to `check_call(...)` -- since we're not actually doing anything with the returned stdout, it _feels_ slightly unnecessary to turn the output into strings that aren't assigned to anything.

On the flip side, there's almost certainly no performance change that can be measured from this and perhaps there's more value in keeping these consistent with the statements where we do care about the output.

integration_tests/test_commands/tests/test_kbenv.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # deploy stack
+        check_output(
+            ['runway', 'deploy', '--tag', 'eks'],
+            cwd=self.stack_path,
+            env=self.env
+        ).decode()
+
+        # create kubeconfig
+        response = check_output(
```
troyready: Assignment of the output here to a variable can be dropped

PR #126, opened 2020-01-09
Narrow module tags integration test role permissions

edgarpoce:
**Goal**
Narrow the IAM permissions of the codebuild project that runs the module tags integration tests

**Implementation**
Added support to customize iam permissions on a test suite basis via yaml files


integration_tests/test_commands/policies.yaml
```diff
@@ -0,0 +1,7 @@
+---
+- Version: "2012-10-17"
+  Statement:
+    - Effect: Deny
```
ITProKyle: This will prevent codebuild from being able to write logs to cloudwatch. Can we add support for no additional policies? that can even be the default.

integration_test_infrastructure/codebuild/common/codebuild_blueprints/iam_policy_builder.py
```diff
@@ -0,0 +1,96 @@
...
+                for entry in entries:
+                    policy = iam.Policy(
+                        PolicyName='inline-policy',
+                        PolicyDocument=entry
+                    )
+                    policies.append(policy)
+            return policies
+        except IOError:
+            print("policies.yaml not found for {0} at {1}".format(test_name, file_path))
+            return []
```
ITProKyle: you can use `path.isfile` to check if the file exists before trying to open so you don't need the `try; except`.

also, use `logging` instead of `print`. `LOGGER` can be initialized at the top of the file with `LOGGER = logging.getLogger(__name__)`

```suggestion
        policies = []
        if path.isfile(file_path):
            with open(file_path, 'r') as stream:
                entries = yaml.safe_load(stream)
                for entry in entries:
                    policy = iam.Policy(
                        PolicyName='inline-policy',
                        PolicyDocument=entry
                    )
                    policies.append(policy)
        else:
            LOGGER.warning('policies.yaml not found for %s at %s', test_name,
                           file_path)
        return policies       
```

integration_test_infrastructure/codebuild/common/codebuild_blueprints/iam_policy_builder.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+"""Module responsible of creating IAM roles."""
+from __future__ import print_function
```
ITProKyle: can be removed

PR #131, opened 2020-01-15
Adding "type" to runway config and abstracting class_path detection.

voodooGQ:
Allows for the use of "type" in runway files to define a module's classification. Runway will first look to see if a class_path has been defined, next if a directory extension is present, next it will see if a "type" is present, and finally if neither of those options produce a class_path variable then it will attempt autodetection by the files contained within.

Because this process is so specialized I moved it to a `ModulePath` class. I'm not a fan of the name I chose here and am looking for thoughts/guidance on a better naming strategy.


src/runway/module/staticsite.py
```diff
@@ -27,7 +27,9 @@ class StaticSite(RunwayModule):
 
     def setup_website_module(self, command):
         """Create stacker configuration for website module."""
-        name = self.options.get('name', self.options.get('path'))
+        original_name = self.options.get('name', self.options.get('path'))
+        # Make sure we remove any `.static` extension
+        name = original_name.split('.')[0]
```
troyready: I can't think of any specific conflicts / edge cases at the moment, but I think it'd be better to codify the exact intent here. Something like:

```
name = self.options.get('name', self.options.get('path'))
directory_suffix = '.static'
if name.endswith(directory_suffix):
    name = name[:-len(directory_suffix)]
```

src/runway/config.py
```diff
@@ -509,7 +526,7 @@ class Config(ConfigComponent):
             ignore_git_branch: true
             tests:
               - name: example
-                type: script
+                type: static
```
ITProKyle: I don't think you meant to change this.

voodooGQ: You would be correct sir, thank you!

src/runway/module_type.py
```diff
@@ -0,0 +1,186 @@
...
+import os
+import sys
+
+from .util import (load_object_from_string)
+
+LOGGER = logging.getLogger('runway')
+
+
+# noqa pylint: disable=too-few-public-methods
+class ModuleType(object):
```
ITProKyle: This should _probably_ be renamed to avoid potential confusion/conflicts. It is the name of a built-in type - https://docs.python.org/3/library/types.html#types.ModuleType. (ran across this working on another project)

It doesn't matter to me what its changed to though. Maybe `RunwayModuleType` if you don't want to get creative.

voodooGQ: I do actually want to get creative on it, but I'm having a hard time defining it since we have "module" defined in a few places already and "type" isn't any better. Whats the feelings on ConfigType? Open to other suggestions as well, `RunwayModuleType` is a bit cumbersome but could be used in a pinch if we can't think of a better one.

ITProKyle: `ConfigType` may be too broad. feel free to use something different then `type` if you have a better idea for that as well. i had picked it as an easy way to infer meaning/intent in the task.

voodooGQ: I agree that "type" is probably the right thing here, it's just too bad these terms intermingle so much. In the interest of getting this pushed out I'm going to go with your suggestion of `RunwayModuleType` and if we think of a better name internally for it down the line we can address it at that point. Thanks for the headsup @ITProKyle

src/runway/config.py
```diff
@@ -60,6 +60,8 @@ class ModuleDefinition(ConfigComponent):  # pylint: disable=too-many-instance-at
     +------------------+-----------------------------------------------+
     | ``.k8s``         | `Kubernetes`_                                 |
     +------------------+-----------------------------------------------+
+    | ``.static``      | :ref:`Static Site<mod-staticsite>`            |
```
troyready: @voodooGQ update to new .web extension

runway/module/staticsite.py
```diff
@@ -28,6 +28,10 @@ class StaticSite(RunwayModule):
     def setup_website_module(self, command):
         """Create stacker configuration for website module."""
         name = self.options.get('name', self.options.get('path'))
+        # Make sure we remove any `.static` extension
+        directory_suffix = '.static'
```
ITProKyle: should this be stripping `.web` instead now?

PR #132, opened 2020-01-18
move embedded stacker to a permanent home

ITProKyle:
The first stage of removing Stacker as a dependency and starting full development for it as a component built into Runway.

With this first step, everything Stacker executed by Runway no longer uses the internal copy. Only the classes/methods used internally will be using it. The follow up to this will be a "shim" so that all uses of Stacker that are executed by Runway will instead use this internal component.


ITProKyle:
closing in favor of #134 which furthers the integration.

PR #137, opened 2020-02-03
revert nesting runway in src directory

ITProKyle:
left these off of CHANGELOG since they are more project maintenance related then actual changes to runway itself

# Added

- a command to Makefile to sync all pipenvs at once (matching the command to update all the lock files at once)
- `mkdir src` to `npm_prep` so it can continue to be used for npm build/publish

# Changed

- `src/runway` to `runway`
- setup.py to remove use of `src`
- setup.py version from `src.runway.__version__` to `runway.__version`
- all Pipfiles to install runway in editable mode from the root of the repo
- developer docs to remove old instructions for how to develop runway and replaced it with a `Getting Started` section that describes how to correctly setup a dev env for runway
- Makefile and travis tests to remove `src` from paths they are excluding
- gitignore to ignore all of `src` instead of a few dirs

# Removed

- path manipulation from docs conf (not needed since runway is installed)


troyready:
Only thing I'm mildly torn on is whether we should still include `PIPENV_VENV_IN_PROJECT=1` in the .env file as well. I'm inclined to do it (a little redundant with the Makefile but maybe appropriately so?), but not set on it.

ITProKyle:
> we should still include PIPENV_VENV_IN_PROJECT=1 in the .env file

I removed it since it requires a manual step for consumption from the .env file. For the pipeline, (which would use it automatically) it doesn't really matter where the venv gets created.

PR #140, opened 2020-02-06
Improvements to AWS Lambda Upload Hook

cmilam87:
## Summary

* support for Pipfile and requirements.txt for lambda functions
* support for restoring packages with docker
* utilizes [docker-py](https://docker-py.readthedocs.io/en/stable)

## Screenshots

image


cmilam87:
**Note**: The integration_test_infrastructure will need to be redeployed to pickup the new tests.

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -163,6 +173,176 @@ def _zip_from_file_patterns(root, includes, excludes, follow_symlinks):
...
+                                        type='bind')
+    pip_cmd = (
+        'python -m pip install -t /var/task -r /var/task/requirements.txt'
+    )
+
+    # TODO figure out what to do with the logs returned here or drop the var
+    _ = docker.containers.run(image=docker_image,
+                              command=['/bin/sh', '-c', pip_cmd],
+                              auto_remove=True,
+                              mounts=[work_dir_mount])
```
ITProKyle: i think this section can be replaced with this to stream the logs. `stream=True` will return a generator we can iterate over to log the response.

```suggestion
    logs = docker.containers.run(image=docker_image,
                                 command=['/bin/sh', '-c', pip_cmd],
                                 auto_remove=True,
                                 mounts=[work_dir_mount],
                                 stream=True)

    for line in logs:
        LOGGER.info(line)
```

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -163,6 +173,176 @@ def _zip_from_file_patterns(root, includes, excludes, follow_symlinks):
...
+    """
+    # TODO stream process output to stdout/stderr
+    LOGGER.info('Creating requirements.txt from Pipfile...')
+    req_path = os.path.join(dest_path, 'requirements.txt')
+    cmd = ['pipenv', 'lock', '--requirements', '--keep-outdated']
+    with open(req_path, 'w') as requirements:
+        pipenv_process = subprocess.Popen(cmd, cwd=package_root,
+                                          stdout=requirements,
+                                          stderr=subprocess.PIPE)
+        _stdout, stderr = pipenv_process.communicate(timeout=timeout)
```
ITProKyle: I _think_ something like this would work here to allow realtime log streaming but would need to be tested.

```suggestion
        pipenv_process = subprocess.Popen(cmd, cwd=package_root,
                                          stdout=requirements,
                                          stderr=subprocess.PIPE)
        for line in pipenv_process.stderr:
            LOGGER.info(line)
        _stdout, stderr = pipenv_process.communicate(timeout=timeout)
```

runway/cfngin/exceptions.py
```diff
@@ -261,6 +275,31 @@ def __init__(self, stack_name, output, *args, **kwargs):
         super(OutputDoesNotExist, self).__init__(message, *args, **kwargs)
 
 
+class PipError(Exception):
+    """Raised when pip returns a non-zero exit code."""
+
+    def __init__(self):
+        """Instantiate class."""
+        # TODO refine message
```
troyready: Let's drop the TODOs here -- seems good enough for now.

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -2,19 +2,29 @@
...
-import os.path
 import stat
+import subprocess
+import sys
 from io import BytesIO as StringIO
+from shutil import copyfile
 from zipfile import ZIP_DEFLATED, ZipFile
 
 import botocore
+import docker as Docker  # noqa: N812
```
troyready: Seem unnecessary to do the docker -> Docker import rename (would prefer to get rid of the flake8 exception)

integration_tests/test_hooks/tests/aws_lambda/blueprints/lambda.py
```diff
@@ -0,0 +1,124 @@
...
+                            Version='2012-10-17',
+                            Statement=[
+                                Statement(
+                                    Action=[
+                                        awacs.logs.CreateLogGroup,
+                                        awacs.logs.CreateLogStream,
+                                        awacs.logs.PutLogEvents
+                                    ],
+                                    Effect=Allow,
+                                    Resource=['arn:aws:logs:*:*:*'],
```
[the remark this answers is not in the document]
ITProKyle: does it even need to deploy a lambda function? we only really need to check that a zip was made, what is in the zip, and that it was placed in s3.

integration_tests/Pipfile
```diff
@@ -4,10 +4,11 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-runway = {editable = true, path = "./.."}
+runway = {editable = true,path = "./.."}
 prettytable = "*"
 colorama = "<=0.3.9,>=0.2.5"
 pexpect = "*"
+boto3 = "*"
```
ITProKyle: `boto3` is a dependency of runway so its not needed here

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -401,7 +592,19 @@ def upload_lambda_functions(context, provider, **kwargs):
             build. Keys correspond to function names, used to derive key
             names for the payload. Each value should itself be a dictionary,
             with the following data:
-
+            **use_pipenv (Optional[bool])**:
+                Will determine if pipenv will be used to generate requirements.txt
+                from an existing Pipfile.
+            **pipenv_timeout (Optional[int])**
+                Time in milliseconds to wait while running pipenv.
```
ITProKyle: `Popen.communicate()` timeout is in seconds, not milliseconds.

> If the process does not terminate after timeout seconds, a TimeoutExpired exception will be raised. Catching this exception and retrying communication will not lose any output.

https://docs.python.org/3.7/library/subprocess.html#subprocess.Popen.communicate

integration_tests/test_hooks/tests/aws_lambda/nondockerize_src/nondockerize.py
```diff
@@ -0,0 +1,22 @@
+"""Test handler."""
+from urllib.reuqest import urlopen, Request
+
+def handler(event, context):
+    """Handle lambda."""
+    req = Request('https://api.github.com')
+    with urlopen(req) as url:
+        f = url.read().decode('utf8')
+        print(f)
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
cmilam87: Not sure why tests_hooks is even in here - it's been moved/refactored into cfngin tests...

cmilam87:
> Were you able to get live output from pipenv & docker working?
> 
> Still missing unit tests.

Wasn't able to get live output working - not sure the best solution.
Which unit tests?

ITProKyle:
> Wasn't able to get live output working - not sure the best solution.

iirc the last we discussed we had found a solution but, because somewhere along the line of how runway/cfngin outputs were getting lost, even what was suppose to be going into `requirements.txt`. have you narrowed down where that was happening?

> Which unit tests?

unit tests for your additions. there are already unit tests for what existed previously in `tests/cfngin/hooks/test_aws_lambda.py`.

ITProKyle:
> live output from pipenv

just dug into this a bit and i believe the issue is due to how Stacker/cfngin locks the UI. either way, its not something that will get done this task. in fact, we arn't getting anything back from the pipenv subprocess to print at the end if we even wanted to. `stderr == None` when run through runway instead of what we were getting running it from it's own file.

ITProKyle:
I was able to get real-time docker logs working

image

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -163,6 +173,195 @@ def _zip_from_file_patterns(root, includes, excludes, follow_symlinks):
...
+    ) as tmpdir:
+        tmp_req = os.path.join(tmpdir, 'requirements.txt')
+        if use_pipenv:
+            _handle_use_pipenv(package_root, tmpdir,
+                               kwargs.get('pipenv_lock_timeout', 300))
+            # dir created when creating a lock file
+            excludes.append('.venv/')
+        for file_name in _find_files(package_root, includes, excludes,
+                                     follow_symlinks):
+            copyfile(os.path.join(package_root, file_name),
```
ITProKyle: this breaks if the package has any directories. I added `lib/__init__.py` to `dockerize_src` to break this for the example in integration_tests

integration_tests/test_hooks/tests/aws_lambda/dockerize_src/dockerize.py
```diff
@@ -0,0 +1,12 @@
+"""Test handler."""
+import requests
```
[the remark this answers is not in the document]
ITProKyle: I swapped the docker/pipenv test to use numpy but, to test the docker component properly (verify binary components from codebuild) we would need to put a bit of work into our codebuild deployment/config for to be able to run docker.

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -322,10 +615,19 @@ def _upload_function(s3_conn, bucket, prefix, name, options, follow_symlinks,
     # absolute path, which is exactly what we want.
     if not os.path.isabs(root):
         root = os.path.abspath(os.path.join(get_config_directory(), root))
-    zip_contents, content_hash = _zip_from_file_patterns(root,
-                                                         includes,
-                                                         excludes,
-                                                         follow_symlinks)
+    if options.get('dockerize_pip'):
```
troyready: @ITProKyle we had discussed leaving this as is, and then changing in the near future when we need to be able to use pip without docker, but I'm still having some reservation about making the change after release.

I'm not wholly against having a different default than the sls plugin, but I'm still not quite convinced that it's worth having it be [different](https://github.com/UnitedIncome/serverless-python-requirements#install). I'd rather we align it (e.g. check here for a Pipfile.lock or requirements.txt, force users to set the docker option if they want to use it, etc) for consistency.

ITProKyle: ok, so you are thinking:

- always try to use pip - don't require docker
  - check for/use `requirements.txt` unless told to use something else
  - check for/use `Pipfile` if the above is not found or explicitly told to do so
- leave docker as is, just don't lock the use of pip behind it

If we're going this route, it may also be worth adding a `python_path` argument to handle `python`/`python3`/etc path differences [sls does similar](https://github.com/UnitedIncome/serverless-python-requirements#customize-python-executable). we can default to using `sys.executable` if this is not provided.

We would also need to include handling for our `pyinstaller` builds. do we disable this if using `pyinstaller`? require `python_path`? include pip and pipenv in the build?

troyready: Agreed on all accounts.

> include pip and pipenv in the build so we can use run-python?
I'll admit the idea crossed my mind but I wasn't sure if I'd be asking for too much :)

ITProKyle: > include pip and pipenv in the build so we can use run-python

Since this will _probably_ be an edge case, think it can slide to a later release so we can get this one out? The other changes will are doable in a day since most of the logic is already there, just needs some rearranging.

troyready: Yeah, sounds good, I agree. I think it'll be a really nice featureset to have but it's seems additive (i.e. won't change anyone's use/configs after it's added & will just enable more uses of it) to me and fine to add in the next release

PR #147, opened 2020-02-11
cfngin persistent graph

ITProKyle:
adapted from cloudtools/stacker#749

> Using a persistent graph enables a Terraform-like experience. When a stack is removed from the local config file, it will be removed from AWS. This will occur for both `build` and `destroy` actions. The `graph` and `diff` commands also take the persistent graph into account (if used).
> 
> This feature requires users **opt-in** to prevent any unexpected results in current environments. This is done by adding a the `persistent_graph_key` directive to any config file. The user is responsible for ensuring that this value is unique per cfngin bucket + namespace.
> 
> The graph is persisted in the existing cfngin bucket (meaning it requires use of a cfngin bucket). No additional resources are required to ensure a low cost and minimally intrusive solution. It is recommended to enable versioning for the bucket to rollback any unintended changes to the object but it is not required and is not actively used by CFNgin. The user is warned each use when it is not enabled and if CFNgin created the bucket when the feature is opted-in, it creates the bucket with versioning enabled.
> 
> The persistent graph object is locked using tags on the S3 object (instead of a DynamoDB table like Terraform) to reduce cost and requirements. When the graph is locked, no other action that acts on the persistent graph (`build`/`destroy`) can be started.
> 
> # What Changed
> 
> - update provider destroy_stack method to have an interactive mode
> - pass region option into context
> - add persist graph to context
>   - context method to get graph
>   - context method to lock graph
>   - context method to unlock graph
>   - context method to put graph
> - add persistent graph exceptions
> - update ensure_s3_bucket
>   - check for recommended bucket settings
>   - applies recommended settings if creating a bucket
> - update plan
>   - class methods to create Step and Graph
>   - remove build_graph
>   - remove build_plan, all logic is now in __init__
>   - add persistent graph management
>   - add more methods for manipulating the graph
> - update base action
>   - add handling for persistent graph
>   - move plan creation to a method that uses classmethods of Graph and Plan
> - update graph action
>   - uses parent class method to create plan
>   - can include persistent graph
> - update diff action
>   - uses parent class method to create plan
>   - can include persistent graph
> - update build action
>   - custom method to create plan that deletes stacks that are in the persistent graph but are no longer in the local graph
> - update destroy action
>   - uses parent class method to create plan
>   - can include persistent graph
> - add persistent_graph_key to config

image
(test name in screenshot was before updating it reflect the test)

### DIFF
--- a/docs/source/cfngin/config.rst
+++ b/docs/source/cfngin/config.rst
@@ -130,7 +130,7 @@ like the below.
 When executing an action that will be modifying the persistent graph
 (build or destroy), the S3 object is *"locked"*. The lock is a tag applied to
 the object at the start of one of these actions. The tag-key is
-**stacker_lock_code** and the tag-value is UUID generated each time a command
+**cfngin_lock_code** and the tag-value is UUID generated each time a command
 is run. In order for CFNgin to lock a persistent graph object, the tag must
 not be present on the object. For CFNgin to act on the graph_ (modify or
 unlock) the value of the tag must match the UUID of the current CFNgin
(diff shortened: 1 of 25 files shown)


ITProKyle:
failing checks pending #145

PR #163, opened 2020-02-15
Add --config option to serverless module

edgarpoce:
Add functionality to customize the name of the sls config file

## Summary

Added small change to include the option
Added small change to the sls documentation

### DIFF
--- a/docs/source/module_configuration/serverless.rst
+++ b/docs/source/module_configuration/serverless.rst
@@ -145,4 +145,4 @@ Name of your configuration file, if other than serverless.yml|.yaml|.js|.json.
         - modules:
             - path: sampleapp.sls
               options:
-                config: sampleapp.yml
+                sls_configfile: sampleapp.yml
--- a/docs/source/module_configuration/serverless.rst
+++ b/docs/source/module_configuration/serverless.rst
@@ -134,3 +134,15 @@ directory is pre-compiled) via the ``skip_npm_ci`` module option:
           - path: myslsproject.sls
             options:
               skip_npm_ci: true
+
+Specifying serverless config file
+---------------------------------
+Name of your configuration file, if other than serverless.yml|.yaml|.js|.json.
+::
+
+    ---
+    deployments:
+        - modules:
+            - path: sampleapp.sls
+              options:
+                config: sampleapp.yml
--- a/runway/module/serverless.py
+++ b/runway/module/serverless.py
@@ -172,6 +172,8 @@ class Serverless(RunwayModule):
         sls_env_file = get_sls_config_file(self.path,
                                            self.context.env_name,
                                            self.context.env_region)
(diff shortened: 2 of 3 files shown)


runway/module/serverless.py
```diff
@@ -172,6 +172,8 @@ def run_serverless(self, command='deploy'):
         sls_env_file = get_sls_config_file(self.path,
                                            self.context.env_name,
                                            self.context.env_region)
+        if self.options.get('options', {}).get('config', {}):
+            sls_opts.extend(['--config', self.options.get('options', {}).get('config', {})])
```
troyready: This conditional check should be changed to only add `--config VAL` if the config option is provided

docs/source/module_configuration/serverless.rst
```diff
@@ -134,3 +134,15 @@ directory is pre-compiled) via the ``skip_npm_ci`` module option:
...
+---------------------------------
+Name of your configuration file, if other than serverless.yml|.yaml|.js|.json.
+::
+
+    ---
+    deployments:
+        - modules:
+            - path: sampleapp.sls
+              options:
+                config: sampleapp.yml
```
troyready: (totally my preference here -- I think it makes sense that you started with the identical option name that sls uses) -- can we make the option `configfile` instead? I'd like to to be more explicit.

runway/module/serverless.py
edgarpoce: I think the current implementation matches the behavior you are describing, I tested it locally and the resulting command ignores de --config parameter when it's not specified

docs/source/module_configuration/serverless.rst
edgarpoce: I'll change it, thanks for reviewing!

runway/module/serverless.py
ITProKyle: this looks to be correct its just difficult to read. maybe something like this would be easier?

```suggestion
        if self.options.get('options', {}).get('config'):
            sls_opts.extend(['--config', self.options['options']['config']])
```

docs/source/module_configuration/serverless.rst
ITProKyle: or maybe even `sls_file` to reduce our use of the word *config*?

runway/module/serverless.py
edgarpoce: @ITProKyle , makes sense, thanks for the suggestion!, I'll update it accordingly

ITProKyle:
make sure this is added to the changelog

docs/source/module_configuration/serverless.rst
edgarpoce: what about sls_configfile?, I'll update the PR with sls_configfile for the moment. Just ping me if there's any concern

PR #168, opened 2020-02-20
avoid npm directory creation errors

troyready:
Certain versions of nodejs can have issues with the directory creation in the current symlink setup process. This small change should make it slightly more forgiving.

image


troyready:
> ... or can it be assumed that the intended result was still achieved?
Yeah, that's what I was thinking

PR #172, opened 2020-02-21
pin back version of virualenv and pyinstaller to fix linux error

ITProKyle:
## Summary

Revert some of the changes from #145 to patch the error reported in pyinstaller/pyinstaller#4674

## Why This Is Needed

```
Error loading Python lib '/usr/bin/libpython3.7m.so.1.0': dlopen: /usr/bin/libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
```

### Added

- `distutils` handling for the versions of `pyinstaller` and `virtualenv` we are now using

### Changed

- `pyinstaller==3.5`
- `virtualenv==16.7.9`

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -22,7 +22,7 @@ fi
 # virtualenv 20 was a complete rewrite and includes some breaking changes
 # with how we were handling things like distutils. These version are now pinned
 # to ensure compatability.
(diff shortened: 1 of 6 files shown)


troyready:
Great stuff, made a build and successfully invoked the executable via a symlink

PR #181, opened 2020-03-04
runway => cfngin integration improvements

ITProKyle:
## Summary

Create an API for CFNgin so we can move away from using `subprocess.Popen` to interface with it as a CLI.

## Why This Is Needed

- improved integration between Runway and its internal component
- persevere data types of Runway `parameters` when passing them to CFNgin

### Added

- `runway.cfngin.cfngin.CFNgin` class (can also be imported as `runway.cfngin.CFNgin`)
- methods to Runway and/or CFNgin context objects for boto3 credentials, region, and getting a thread-safe session
- `AWS_ENV_VARS` to utils for a list of env var names
- `runway.cfngin.context.Context.get_session()` method for creating a boto3 session using the correct AWS credentials
- `runway.cfngin.context.Context.config_path` attribute to record the location of the loaded config file
- some terminology documentation for runway
- `environment` and `region` as _common parameters_ for cloudformation modules so they do not need to be defined

### Changed

- `get_session` can now accept AWS credentials when creating a thread-safe session
- cloudformation module logic is now primarily in the `CFNgin` class
- Runway environment variable options to pull from `self.env_var` instead of `os.environ`
- some unit tests are now `pytest` instead of `unittest`
- deprecated the `run-stacker` command
- deprecated the use `get_session` directly when credentials are in environment variables
- deprecated `runway.cfngin.util.get_config_directory()` which was only used for the aws_lambda hook.
- deprecated Stacker CLI components
- cloudformation module documentation to focus more on the use of `parameters` instead of an environment file


docs/source/module_configuration/cloudformation.rst
```diff
@@ -1,108 +1,142 @@
...
 .. _mod-cfn:
 
 CloudFormation
 ==============
 
-CloudFormation modules are managed by 2 files:
+CloudFormation modules are processed using CFNgin.
+There are two files that can be used for configuration:
+
+- a YAML `CFNgin Config File`_
```
troyready: For any of these touchpoints (places in the docs where initial/basic users of Runway are looking for info on defining stacks & params, etc), I'd really like to demphasize the CFNgin name. Specifically, it'd be nice if we could avoid having people ask questions about runway/cfngin, and just have them ask about "runway".

So for this specific place, something like a "YAML Runway CloudFormation Config File" I _think_ would be better for that goal.

ITProKyle: makes sense - just a bit of a mouth full. I'll go back through a reword this page tomorrow (or later tonight if I get a chance).

PR #184, opened 2020-03-06
ssm lookup

ITProKyle:
## Summary

A feature rich SSM parameter store lookup that works for both Runway and CFNgin config files.

### Added

- `ssm` lookup usable in Runway and CFNgin config files
- `get_session` method to Runway context for parity with CFNgin context
- `format_results` and `load` methods to `LookupHandler` base class
- `staticsite_acmcert_arn` option

### Changed

- `LookupHandler` base class string transform now converts dicts into json escaped strings and supports variable indent levels
- `LookupHandler` base class string transform now wraps `bool` in double quotes
- existing Runway lookups now use `format_results` before returning a value
- deprecate `ssmstore` lookup
- deprecate `staticsite_acmcert_ssm_param` option
- deprecate `terraform_backend_ssm_params` option


ITProKyle:
@troyready - not seeing any other uses of the deprecated options (none in `templates/`). updated a place I missed for `ssmstore` though

image
image

troyready:
@ITProKyle you're right, I was thinking about the CFN lookup!

PR #185, opened 2020-03-07
add editorconfig file

troyready:
## Summary

Adds an editorconfig file

## Why This Is Needed

Should help developers commit new changes with our desired options, without having to wait for CI or PR reviews to give feedback.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


troyready:
> Would it be worth adding indents to this as well?

Yeah, I'd agree. Was too lazy here to start it but it'd be nice.

PR #192, opened 2020-03-12
add static site IAM role permissions boundary support

troyready:
## Summary

Add static site IAM role permissions boundary support

## Why This Is Needed

Deployment scenarios that dictate more limited IAM permissions in an account.


troyready:
Forgot to mention: also dropped the legacy ssm param option from the docs

PR #193, opened 2020-03-13
[Bugfix] Lambda Execution Role Permissions Scope

voodooGQ:
## Summary

**Note: This PR has https://github.com/onicagroup/runway/pull/190 rebased in as there are some shared changes. This PR will be rebased against master once that has been approved.**

Fixes an issue where the Lambda Execution roles were extraneously scoped and provided to many permissions. The pattern presented here aims to keep the interface minimal while addressing the root of the issued by passing in Resource handlers on a function by function basis during the initial call.

## Why This Is Needed

Right thing to do. Make sure we're as narrowly scoped as possible

## What Changed

- Moved the `add_lambda_execution_role` function to the base `staticsite` template
- Configured the `add_lambda_execution_role` function to accept the resource name we're looking to target for that specific function

### Fixed

More narrowly scoped Lambda permissions

### DIFF
--- a/runway/blueprints/staticsite/auth_at_edge.py
+++ b/runway/blueprints/staticsite/auth_at_edge.py
@@ -8,9 +8,9 @@ import logging
 from typing import Any, Dict, List  # pylint: disable=unused-import
 
 import awacs.s3
+import awacs.logs
 from awacs.aws import Allow, Principal, Statement
-from awacs.helpers.trust import make_simple_assume_policy
-from troposphere import (Join, NoValue,  # noqa pylint: disable=unused-import
+from troposphere import (AccountId, Join, NoValue, StackName,  # noqa pylint: disable=unused-import
                          awslambda, cloudfront, iam, s3)
 
 from .staticsite import StaticSite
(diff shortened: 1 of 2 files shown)


runway/blueprints/staticsite/auth_at_edge.py
```diff
@@ -92,47 +92,67 @@ def create_template(self):
...
         check_auth_lambda = self.get_auth_at_edge_lambda_and_ver(
             'CheckAuth',
             'Check Authorization information for request',
             'check_auth',
-            lambda_execution_role
+            self.add_lambda_execution_role(
+                'CheckAuthLambdaExecutionRole',
+                Join(':', ['arn:aws:logs:*',
+                           AccountId,
+                           'log-group:/aws/lambda/*-CheckAuth-*:*'])
```
troyready: (thoughts here will apply to each of the functions)

1) Should probably include the [StackName](https://github.com/cloudtools/troposphere/blob/2.6.0/troposphere/__init__.py#L518) psuedoparam as a prefix here before the resource id (`CheckAuth` in this case) to keep it narrowed to its namespace
2) Would be good to explicitly link the function resource name (4 lines above, the `'CheckAuthLambdaExecutionRole'` string) to its use here (i.e. make it a variable and reference the variable)
3) For normal lambda function operation, it needs to start with `/aws/lambda/FUNCTIONAME` and then the lambda@edge executions need `/aws/lambda/REGION.FUNCTIONAME` -- to really go all out here (and we're already in for a penny ....) I think we'll want to provide the resource of `/aws/lambda/StackName-ResourceName-*` & `/aws/lambda/*.StackName-ResourceName-*`
4) Not a separate change from what was noted in point 3 ^, but just to highlight something implied by it: the CreateLogGroup & CreateLogStream permissions apply to log-group ARN (e.g. `/aws/lambda/StackName-FunctionResource-*`), and the PutLogEvents permissions should apply to the log-stream ARNs (`/aws/lambda/StackName-FunctionResource-*:log-stream:*`). So the ARN can't be `...log-group:/aws/lambda/*-CheckAuth-*:*` for all 3. Easiest option to just have a single resource ARN that will cover both is to omit the `:*` at the end

runway/blueprints/staticsite/auth_at_edge.py
```diff
@@ -92,47 +92,67 @@ def create_template(self):
...
         # Auth@Edge Lambdas
-        lambda_execution_role = self.add_lambda_execution_role()
         check_auth_lambda = self.get_auth_at_edge_lambda_and_ver(
             'CheckAuth',
             'Check Authorization information for request',
             'check_auth',
-            lambda_execution_role
+            self.add_lambda_execution_role(
+                'CheckAuthLambdaExecutionRole',
+                Join(':', ['arn:aws:logs:*',
```
troyready: Another non-functional change here (only relevant if someone copies the code to one of the isolated regions) -- should swap `aws` here for [Partition](https://github.com/cloudtools/troposphere/blob/2.6.0/troposphere/__init__.py#L515)

voodooGQ:
@troyready, we had lots of offline discussion. Here's the generated policy with the current change set. Take a look over when you have a chance and make sure we're addressing everything necessary: 
image

PR #194, opened 2020-03-13
Auth@Edge User Pool Creation

voodooGQ:
## Summary

**Note: This PR is dependent on two others being implemented that are in review state: https://github.com/onicagroup/runway/pull/190 && https://github.com/onicagroup/runway/pull/193**

This introduces the ability for the user to generate a basic User Pool with their Auth@Edge build.

## Why This Is Needed

Gives more flexibility to the User, allowing them to spin up an entire Auth@Edge suite without the need to supply manually created elements.

## What Changed

Made references in all pre_hooks from the data being supplied by `aae_create_user_pool_id` pre_hook.

### Added

- Ability to create a User Pool
- Domain Updater check to remove the generated domain prior to User Pool Deletion

### DIFF
--- a/docs/source/module_configuration/staticsite.rst
+++ b/docs/source/module_configuration/staticsite.rst
@@ -78,8 +78,17 @@ Parameters
 
     staticsite_auth_at_edge: true
 
+**staticsite_create_user_pool (Optional[bool])**
+  Creates a User Pool for the Auth@Edge configuration. Either this or ``staticsite_user_pool_arn`` are required when ``staticsite_auth_at_edge`` is ``true``.
+
+  Example:
+
+  .. code-block:: yaml
+
+    staticsite_create_user_pool: true
+
 **staticsite_user_pool_arn (Optional[str])**
-  Required if ``staticsite_auth_at_edge`` is ``true``. A pre-existing Cognito User Pool is required for user authentication.
+  A pre-existing Cognito User Pool is required for user authentication. Either this or ``staticsite_create_user_pool`` are required when ``staticsite_auth_at_edge`` is ``true``.
 
   Example:
 
@@ -343,7 +352,7 @@ Here is how the solution works:
 6. After passing all of the verification steps, `Lambda@Edge` strips out the Authorization header and allows the request to pass through to designated origin for CloudFront. In this case, the origin is the private content Amazon S3 bucket.
 7. After receiving response from the origin S3 bucket, CloudFront sends the response back to the browser. The browser displays the data from the returned response.
 
-An example of a `Auth@Edge`_ static site configuration is as follows. All listed options are required:
+An example of a `Auth@Edge`_ static site configuration is as follows:
 
 .. code-block:: yaml
 
@@ -362,5 +371,5 @@ An example of a `Auth@Edge`_ static site configuration is as follows. All listed
         - us-east-1
 
 The `Auth@Edge`_ functionality uses your existing Cognito User Pool (optionally configured
-with federated identity providers). A user pool app client will be automatically created
-within the pool for the application's use.
+with federated identity providers) or can create one for you with the ``staticsite_create_user_pool`` option.
+A user pool app client will be automatically created within the pool for the application's use.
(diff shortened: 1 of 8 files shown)


docs/source/module_configuration/staticsite.rst
```diff
@@ -78,8 +78,17 @@ Parameters
...
+
+  Example:
+
+  .. code-block:: yaml
+
+    staticsite_create_user_pool: true
+
 **staticsite_user_pool_arn (Optional[str])**
-  Required if ``staticsite_auth_at_edge`` is ``true``. A pre-existing Cognito User Pool is required for user authentication.
+  A pre-existing Cognito User Pool is required for user authentication. Either this or ``staticsite_user_pool_arn`` are required when ``staticsite_auth_at_edge`` is ``true``.
```
troyready: Should be "Either this or ``staticsite_create_user_pool``"

fwiw, you could probably convince me that the need to create a user pool is implicit when an ARN isn't provided, i.e. we could drop the `staticsite_create_user_pool` option and just make it a check for `staticsite_user_pool_arn` being specified. Really not sure what's more logical. I generally prefer explicit configuration, but it can get unwieldy

runway/blueprints/staticsite/dependencies.py
```diff
@@ -110,17 +115,30 @@ def create_template(self):
...
         ))
 
+        callbacks = self.context.hook_data['aae_callback_url_retriever']['callback_urls']
+
         if variables['AuthAtEdge']:
+            user_pool_id = variables['UserPoolId']
+
+            if variables['CreateUserPool']:
+                user_pool = template.add_resource(
+                    cognito.UserPool("AuthAtEdgeUserPool")
```
troyready: (entirely style & preference) can we make the resource just "UserPool" -- there won't be any other user pools in the stack & its nice to avoid prefixing all the resources with the general stack concept.

runway/blueprints/staticsite/dependencies.py
```diff
@@ -110,17 +115,30 @@ def create_template(self):
...
+                )
+
+                user_pool_id = user_pool.ref()
+
+                template.add_output(Output(
+                    'AuthAtEdgeUserPool',
```
troyready: Also just style/preference -- let's call this `UserPoolId`

i.e. appending the value type, so its clear/unique compared to others we might add in the future like `UserPoolArn`. Nothing would be wrong about just having `UserPool` & `UserPoolArn`, but my preference is just to start explicit

runway/blueprints/staticsite/dependencies.py
voodooGQ: Should we do the same with the Client? It'd take a little refactoring but not much.

docs/source/module_configuration/staticsite.rst
voodooGQ: Yeah I thought about that for a while too and was on the fence. I ultimately went with the explicit because I felt it was necessary to make the user fully aware of the decision. If they don't realize they have to create/supply a user pool they may not understand how to populate it with user values.

I'm in the same boat though and could be convinced otherwise.

runway/blueprints/staticsite/dependencies.py
troyready: That is interesting -- we can't update the client now (stacks are already deployed with that resourceid).

I'm not sure it's a big deal to have them be inconsistent, but it's fine leaving this as AuthAtEdgeUserPool if it seems more consistent.

voodooGQ: Ok, I'll leave as is just to keep the naming consistent.

runway/blueprints/staticsite/dependencies.py
voodooGQ: In regards to above should we make this `AuthAtEdgeUserPoolId`?

troyready:
Should be ready to de-draft now

voodooGQ:
@troyready Ready for review

PR #201, opened 2020-03-20
StaticSite Cleanup State Machine

voodooGQ:
## Summary

In the process of creating a Static Site with a Lambda@Edge function and then subsequently attempting to delete it a "Replicated Function" will be left behind https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-edge-delete-replicas.html. As the article outlines these cannot be deleted until CloudFront has propagated its deletion. 

Runway will attempt to delete the function for about an hour and finally "give up" leaving these functions orphaned whilst making the user experience less than desirable.

This PR launches a StateMachine designed to keep attempting deletion of these functions until successful. Once they have been deleted the State Machine will then delete the Stack it's a part of to "self-destruct"

## Why This Is Needed
1) Better User Experience
2) No orphaned resources

### Added

1) Replicated Function Remover State Machine
2) Corresponding Lambdas/Roles
3) Logic to ensure cleanup is only run when necessary and doesn't get deleted during a normal stack destroy

### DIFF
--- a/runway/blueprints/staticsite/cleanup.py
+++ b/runway/blueprints/staticsite/cleanup.py
@@ -142,7 +142,6 @@ class Cleanup(Blueprint):
 
         return res
 
-    # TODO: Swap this with paramterized_codec
     def _get_self_destruct(self, replicated_lambda_remover):
         # type: (Dict[str, Union[awslambda.Function, Any]]) -> Dict[str, Any]
         res = {}
--- a/runway/blueprints/staticsite/cleanup.py
+++ b/runway/blueprints/staticsite/cleanup.py
@@ -1,0 +1,415 @@
+#!/usr/bin/env python
+"""Cleanup Stack.
+
+Responsible for cleaning up the remaining orphaned resources created
+by the primary stack.
+"""
+from __future__ import print_function
+
+import json
+import os
+import logging
+from typing import Any, Dict, Union  # pylint: disable=unused-import
+
+import awacs.awslambda
+import awacs.cloudformation
+import awacs.iam
+import awacs.logs
+
+from awacs.aws import Allow, PolicyDocument, Statement
+from awacs.helpers.trust import make_simple_assume_policy
+from troposphere import (AccountId, Join, Output, Partition, Region, StackName, Sub,
+                         awslambda, iam, stepfunctions)
+
+from runway.cfngin.blueprints.base import Blueprint
+from runway.cfngin.util import read_value_from_path
+
+LOGGER = logging.getLogger('runway')
+
+
+class Cleanup(Blueprint):
+    """Cleanup Stack."""
+
+    VARIABLES = {
+        'DisableCloudFront': {'type': bool,
+                              'default': False,
+                              'description': 'Whether to disable CF'},
+        'function_arns': {'type': list,
+                          'default': [],
+                          'description': 'List of function ARNs that need to '
+                                         'be removed.'},
+        'stack_name': {'type': str,
+                       'default': '',
+                       'description': 'The name of the current stack'}
+    }
+
+    @property
+    def cf_enabled(self):
+        # type: () -> bool
+        """CloudFront enabled conditional."""
+        return not self.get_variables().get('DisableCloudFront', False)
+
+    def create_template(self):
+        # type: () -> None
+        """Create template (main function called by Stacker)."""
+        self.template.set_version('2010-09-09')
+        self.template.set_description('Static Website Cleanup - StateMachine')
+        if not self.cf_enabled:
+            return
+
+        self._get_replicated_lambda_state_machine()
+
+    def _get_replicated_lambda_remover_lambda(self):
+        # type: () -> Dict[str, Any]
+        res = {}
+        res['role'] = self.template.add_resource(
+            iam.Role(
+                'ReplicatedLambdaRemoverRole',
+                AssumeRolePolicyDocument=make_simple_assume_policy('lambda.amazonaws.com'),
+                Policies=[
+                    iam.Policy(
+                        PolicyName="LambdaLogCreation",
+                        PolicyDocument=PolicyDocument(
+                            Version='2012-10-17',
+                            Statement=[
+                                Statement(
+                                    Action=[awacs.logs.CreateLogGroup,
+                                            awacs.logs.CreateLogStream,
+                                            awacs.logs.PutLogEvents],
+                                    Effect=Allow,
+                                    Resource=[
+                                        Join('', [
+                                            'arn:',
+                                            Partition,
+                                            ':logs:*:',
+                                            AccountId,
+                                            ':log-group:/aws/lambda/',
+                                            StackName,
+                                            '-ReplicatedLambdaRemover-*'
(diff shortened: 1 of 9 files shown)


runway/blueprints/staticsite/staticsite.py
```diff
@@ -581,6 +594,17 @@ def _get_index_rewrite_role_function_and_version(self):
         )
         return res
 
+    def _get_code_template(self, relative_template_location):  # pylint: disable=no-self-use
```
troyready: Can we drop this function in favor the native cfngin parameterized lookup? Should be functionally about the same (returning a Join CFN object instead that can be used where the string is used).

runway/blueprints/staticsite/templates/replicated_lambda_remover.template.py
```diff
@@ -0,0 +1,67 @@
...
+
+    deletions = []
+
+    try:
+        # Loop through all arns passed in
+        for _ in range(0, len(arns)):
+            try:
+                deletions = [
+                    lambda_client.delete_function(FunctionName=arn) for arn in arns
+                ]
```
troyready: I think this loop logic can be reduced slightly, something like:

```
        deletions = []
        for arn in arns:
            try:
                deletions.append(lambda_client.delete_function(FunctionName=arn))
```

runway/blueprints/staticsite/templates/replicated_lambda_remover.template.py
```diff
@@ -0,0 +1,67 @@
...
+            'FunctionArns': arns,
+            'deleted': True,
+            'deletions': deletions,
+            'status': 'Delete Successful'
+        }
+    # Something went wrong, more than likely the Replicated Function is
+    # not currently able to be deleted.
+    # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-edge-delete-replicas.html
+    # Try again
+    except Exception:  # pylint: disable=broad-except
```
troyready: Would be nice to first catch whatever FunctionDoesNotExist exceptions and no-op on them

runway/blueprints/staticsite/cleanup.py
```diff
@@ -0,0 +1,376 @@
...
+                        PolicyName="DeleteLambda",
+                        PolicyDocument=PolicyDocument(
+                            Version="2012-10-17",
+                            Statement=[
+                                Statement(
+                                    Action=[awacs.awslambda.DeleteFunction],
+                                    Effect=Allow,
+                                    Resource=self.get_variables()['function_arns']
+                                )
+                            ]
```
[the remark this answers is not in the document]
voodooGQ: This was an intentional omition. Didn't want the stack to leave a trace behind (and didn't want to have to delete the logs as well). Let me know if you'd prefer them logging, can change tact.

runway/blueprints/staticsite/templates/replicated_lambda_remover.template.py
voodooGQ: From other PRs I've had I've been told that using comprehensions is the more "pythonic" way of handling these loops. I'm happy to change this if you'd like, but I'm beginning to be unsure in that case when the right time to use them is? Would you be able to give me some pointers in that regard so I make the correct decisions going forward?

runway/blueprints/staticsite/templates/replicated_lambda_remover.template.py
voodooGQ: That's handled further up in the Loop. If the Resource isn't found it continues the loop and moves on to the next. That way if the function has been previously deleted we're not killing the process run: 
image

runway/blueprints/staticsite/cleanup.py
```diff
@@ -0,0 +1,376 @@
...
+    CLEANUP_VARIABLES = {
+        'function_arns': {'type': list,
+                          'default': [],
+                          'description': 'List of function ARNs that need to '
+                                         'be removed.'}
+    }
+
+    VARIABLES = {}  # type: Dict[str, Dict[str, Any]]
+
+    def __init__(self,
```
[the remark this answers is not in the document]
voodooGQ: > I think they're unique enough here to warrant just being in this stack.

I may be misunderstanding what you're trying to convey here. The reason this is being handled like this is because they are unique to this stack. If I add them in as part of the main variables then I'd have to either copy over all the variables from StaticSite (which isn't necessary since we're inheriting) or feed all of the variables into each one of the child templates.

runway/blueprints/staticsite/staticsite.py
voodooGQ: Are you talking about this? https://docs.onica.com/projects/runway/en/latest/apidocs/runway.cfngin.lookups.handlers.file.html#runway.cfngin.lookups.handlers.file.parameterized_codec

runway/blueprints/staticsite/cleanup.py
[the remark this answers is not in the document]
troyready: Put a different way: this stack doesn't seem related enough to the StaticSite to make it worth being a subclass of StaticSite. My preference would be to just make it a subclass of Blueprint and add the `stack_name` parameter.

runway/blueprints/staticsite/cleanup.py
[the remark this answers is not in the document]
troyready: Reasonable thought, but we don't have anything else in place to clean up the other function logs so one more LogGroup shouldn't be a problem (and possibly a nice to have for any future troubleshooting)

runway/blueprints/staticsite/staticsite.py
troyready: Yes

runway/blueprints/staticsite/templates/replicated_lambda_remover.template.py
troyready: It _feels_ like there should be a way to do this without both the `for` loop and the list comprehension (intent isn't to try the deletions `len(arns) * len(arns)` times, right?)

voodooGQ: D'oh. Ok I see what you're saying now. I'll get this fixed.

runway/blueprints/staticsite/cleanup.py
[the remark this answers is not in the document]
voodooGQ: Updated

runway/blueprints/staticsite/cleanup.py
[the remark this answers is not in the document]
voodooGQ: Updated

runway/blueprints/staticsite/templates/replicated_lambda_remover.template.py
voodooGQ: Updated... finally

runway/blueprints/staticsite/staticsite.py
voodooGQ: Updated to use `read_value_from_path`

PR #212, opened 2020-04-01
Fix config error in example snippet for auth@edge

sagardesai0094:
## Summary

Fixed an error in the example snippet shown in the documentation for Auth@Edge. The example was using environment subsections under the parameters key, which does not work. The updated example uses a variables section in the runway file, and look ups in the parameters section.


sagardesai0094:
Github created the branch automatically so the name doesn't follow your convention. If you need me to reopen, let me know.

PR #250, opened 2020-04-21
action to update urlshortener when a new version is published to s3

ITProKyle:
**The IAM user used for publishing to S3 will need its permissions expanded to include the ability to update the DynamoDB table. The permissions change is included in the PR but, needs to be deployed once approved/merged.**

## Summary

Update the `oni.ca` url shortener with a new version and latest entry when a new version is published to s3.

### Added

- script with tests
- job to the `cicd` workflow to run the script

## Screenshots

The script has been tested but not the action. It will require a new release to run or will need to be setup in a different repo to test.

image

image


troyready:
The IAM user is created via the CDK project at https://bitbucket.org/corpinfo/runway-deployment-support -- probably should just move it in here (in this PR) and then we can retire that old repo?

PR #253, opened 2020-04-23
support an AWS SSO profile as the initial credential source 

### LINKED ISSUE #251: [REQUEST] support an AWS SSO profile as the initial credential source
**Is your feature request related to a problem? Please describe.**

When using AWS SSO profiles configured using `awscli>=2.0.0` and selected with `export AWS_PROFILE=<name>`, the credentials cannot be loaded.

This means that in order to use Runway with an AWS SSO profile, the credentials would need to be collected and exported as environment variables or writing the resolved credentials to a file **before** executing Runway. Some users have created a script that handles this and would need to be run before Runway such as [this one](https://gist.github.com/sgtoj/af0ed637b1cc7e869b21a62ef56af5ac)

Other IaC tools such as serverless are experiencing similar issues.

**Describe the solution you'd like**

Embed a dev build of botocore that supports AWS SSO profiles or, patch a AWS SSO credential provider into the general release of botocore.

When Runway is executed, if a `AWS_PROFILE` is set, use the embedded or patched botocore to retrieve the credentials and push them to the environment/`context.env_vars` (replacing `AWS_PROFILE`). This would enable all IaC tools that Runway wraps to take advantage of the AWS SSO credentials.

ITProKyle:
Closes #251 

## Why This Is Needed

The publicly released version of botocore does not support using credentials from an AWS SSO profile configure with awscli>=2.0.0.

This means that if the environment variable `AWS_PROFILE` is set to a profile name that was configured with the command `aws configure sso --profile <name>`, anything that uses botocore to load credentials (like boto3) will raise `NoCredentialsError`.

This issue is not limited to botocore/boto3, it also effects current releases Serverless Framework and Terraform.

### Added

- extension of `botocore.session.Session` that includes the assets required to use AWS SSO profiles as provided in awscli==2.0.0.dev13 (unpublished)

### Changed

- `get_session` passes the extended `botocore.session.Session` session to `boto3.Session` so that it can make use of AWS SSO profiles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- support for AWS SSO profile as the initial credential source
 
 ## [1.7.2] - 2020-04-21
 ### Fixed
--- a/runway/aws_sso_botocore/__init__.py
+++ b/runway/aws_sso_botocore/__init__.py
@@ -1,0 +1,22 @@
+"""Support for AWS SSO.
+
+IMPORTANT: This will be removed upon botocore/boto3 officially supporting
+authentication with AWS SSO profiles.
+
+The code in this directory is taken from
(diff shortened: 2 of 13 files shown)


troyready:
Looks great to me at a glance, nice work.

PR #254, opened 2020-04-24
add args option tf modules, improve option handling with dedicated classes

### LINKED ISSUE #252: [REQUEST] Support for terraform's `parallelism`
**Is your feature request related to a problem? Please describe.**
While parallelism is an advanced feature of terraform it does have it's uses and I'd like to be able to speed up some of the deployments.

**Describe the solution you'd like**
A key specified in the module definition stating parallelism.
```
deployments:
  - modules:
      - name: example
        path: dynamic-uat.tf
        parallelism: 10
```
I'm not tied to this format, there may be better ways such as global.

**Describe alternatives you've considered**
I can manually run terraform and use parallelism.

ITProKyle:
## Summary

Improved Terraform module option handling and support for passing CLI arguments to Terraform commands (`apply`, `init`, and `plan`).

### Added

- `ModuleOptions` base class for module option parsing/handling
- `TerraformOptions` and `TerraformBackendConfig` to simplify parsing, working with, and testing Terraform module options
- support for `args` to be passed to the Terraform module which is passed to a corresponding command while Runway is processing the module (resolves #252)

### Removed

- the following were removed as the logic was moved into the objects that were added
	- `runway.module.terraform.create_config_backend_options`
	- `runway.module.terraform.get_backend_init_list`
	- `runway.module.terraform.gen_backend_tfvars_files`
	- `runway.module.terraform.get_backend_tfvars_files`
	- `runway.module.terraform.get_module_defined_tf_var`


ITProKyle:
While I don't think anything here should be broken, the lack of tests prior to the change and my lack of experience with Terraform should warrant some basic, manual testing by someone more expertise with Terraform.

If there are any potential combinations of options I missed from the new tests that where implemented, please let me know. I tried to capture as many as I could think of but may have overlooked some older format.

PR #337, opened 2020-05-21
Fix npm subprocess execution on Ubuntu 20.04

troyready:
Linux/macOS execution has traditionally required annoying extra quoting.
Now, with Python 3.8 the old process generates an error:

```
INFO:runway:Running sls deploy on backend.sls ("npx -c 'sls deploy -r us-east-1 --stage prod'")
/bin/sh: 1: sls deploy -r us-east-1 --stage prod: not found
```

Conditionally dropping the extra quotes fixes the error.

### DIFF
--- a/runway/module/__init__.py
+++ b/runway/module/__init__.py
@@ -21,10 +21,9 @@ NPX_BIN = 'npx.cmd' if platform.system().lower() == 'windows' else 'npx'
 
 def format_npm_command_for_logging(command):
     """Convert npm command list to string for display to user."""
-    if platform.system().lower() == 'windows':
-        if command[0] == 'npx.cmd' and command[1] == '-c':
-            return "npx.cmd -c \"%s\"" % " ".join(command[2:])
-        return ' '.join(command)
+    if platform.system().lower() == 'windows' and (
+            command[0] == 'npx.cmd' and command[1] == '-c'):
+        return "npx.cmd -c \"%s\"" % " ".join(command[2:])
     return ' '.join(command)
 
 
--- a/runway/module/__init__.py
+++ b/runway/module/__init__.py
@@ -1,5 +1,4 @@
 """Runway module module."""
-from distutils.version import LooseVersion  # noqa pylint: disable=import-error,no-name-in-module
 import logging
 import os
 import platform
@@ -25,10 +24,8 @@ def format_npm_command_for_logging(command):
     if platform.system().lower() == 'windows':
         if command[0] == 'npx.cmd' and command[1] == '-c':
             return "npx.cmd -c \"%s\"" % " ".join(command[2:])
-        return " ".join(command)
-    # Strip out redundant npx quotes not needed when executing the command
-    # directly
-    return " ".join(command).replace('\'\'', '\'')
+        return ' '.join(command)
+    return ' '.join(command)
 
 
 def generate_node_command(command, command_opts, path):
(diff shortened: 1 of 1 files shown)


runway/module/__init__.py
```diff
@@ -39,6 +40,15 @@ def generate_node_command(command, command_opts, path):
             cmd_list = [NPX_BIN,
                         '-c',
                         "%s %s" % (command, ' '.join(command_opts))]
+        # Python 3.8+ removes the silly need for redundant quoting
+        # https://docs.python.org/3/whatsnew/3.8.html#optimizations
+        elif platform.libc_ver()[1] and (
+                LooseVersion(platform.libc_ver()[1]) >= LooseVersion('2.24')) and (
+                    LooseVersion('.'.join([str(i) for i in sys.version_info[0:2]])) >=
+                    LooseVersion('3.8')):
```
ITProKyle, line 48: this can be simplified to

```suggestion
                    sys.version_info >= (3, 8)):
```

runway/module/__init__.py
```diff
@@ -39,6 +40,15 @@ def generate_node_command(command, command_opts, path):
             cmd_list = [NPX_BIN,
                         '-c',
                         "%s %s" % (command, ' '.join(command_opts))]
+        # Python 3.8+ removes the silly need for redundant quoting
+        # https://docs.python.org/3/whatsnew/3.8.html#optimizations
+        elif platform.libc_ver()[1] and (
+                LooseVersion(platform.libc_ver()[1]) >= LooseVersion('2.24')) and (
```
ITProKyle, line 46: the linked article states that this effects macOS and Linux but this condition will never be truthy on macOS.

```shell
$ python -c "import platform; print(platform.libc_ver())"
('', '')
```

troyready: It didn't occur to me yesterday, after having some trouble parsing the changelog, to check [the source](https://github.com/python/cpython/blob/v3.8.0/Lib/subprocess.py#L634-L678) to figure it out (I instead did dumb stuff like test on Alpine).

Can you confirm the change in behavior on macOS @ITProKyle ? Don't need to test the whole thing in runway with serverless, just trying to call subprocess on `npx -c 'foo bar'` on 3.8 vs <3.8. I could do it _but lazy_.

ITProKyle: tested with this script on macOS 10.15.4

```python
import subprocess

subprocess.check_call(['npx', '-c', "sls --version"])
subprocess.check_call(['npx', '-c', "''sls --version''"])
subprocess.check_call(['npx', '-c', "'sls --version'"])
```

with python version 2.7.15, 3.7.6, 3.8.1, and 3.8.2. all of which choke on the third `.check_call()` with a similar error.

```
/bin/sh: sls --version: command not found
Traceback (most recent call last):
  File "test.py", line 5, in <module>
    subprocess.check_call(['npx', '-c', "'sls --version'"])
  File "/Users/kyle/.pyenv/versions/3.8.2/Python.framework/Versions/3.8/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['npx', '-c', "'sls --version'"]' returned non-zero exit status 127.
```

troyready:
Edited title to reflect testing by Kyle -- doesn't seem to affect python 3.8 on ubuntu 18.04.

Going to next investigate simplifying the subprocess logic on all platforms.

troyready:
@ITProKyle hitting this again today reminded me that I never followed up on it.

I don't want to play fast and loose with changes, but I don't think we ever encountered a platform/situation where `subprocess.check_call(['npx', '-c', "sls --version"])` failed, right? Possible the original logic was just overly and unnecessarily cautious?

Given that we have CodeBuild testing sls execution, I'm inclined to just simplify it and remove all the quoting -- any objections?

ITProKyle:
>  but I don't think we ever encountered a platform/situation where `subprocess.check_call(['npx', '-c', "sls --version"])` failed, right

I believe this is correct.

> Given that we have CodeBuild testing sls execution

its only testing on one OS though. it may be worth holding off on merging the change till after integration tests are able to be run with GitHub Actions for more OS coverage.

troyready:
I wasn't able to find any instance (macOS, Ubuntu, Amazon Linux & Ubuntu on CodeBuild) that was a problem with the simplified (non-single-quotes) logic, so I'm going to assume it's original inclusion was overly-cautious and unnecessary.

I'd wait to merge it in, but I strongly suspect any failure cases won't be found with any additional simple testing. If there's something wrong with this approach, I'm not sure if we'll be able to predict it beforehand (we'll just need to add a fix & test if it comes up)

PR #341, opened 2020-05-22
add `future.strict_environment` config option

### LINKED ISSUE #333: [REQUEST] deployment/module.environments should be strict
As show, this will break legacy uses of `environments`. This would need to be released in a major release or, made toggleable with a default of `false` until the next major release. It should be a top-level config option.

# Example

runway.yml

```yaml
deployments:
  - modules:
      - path: sampleapp.cfn
        parameters:
          key: val
    environments:
      example: 0000/us-east-1
    regions:
      - us-east-1
      - us-west-2
```

## Expectation

- `DEPLOY_ENVIRONMENT=example, accountId=0000, region=us-east-1` deploy the module
- `DEPLOY_ENVIRONMENT=example, accountId=0000, region=us-west-2` always skip the module because the region does not match
- `DEPLOY_ENVIRONMENT=example, accountId=1111, region=us-east-1` always skip the module because the accountId does not match
- `DEPLOY_ENVIRONMENT=prod, accountId=0000, region=us-east-1` always skip the module because the environment is not defined
- `DEPLOY_ENVIRONMENT=prod, accountId=0000, region=us-west-2` always skip the module because the environment is not defined

If `environments` is not defined for an deployment/module, don't skip, let the module class determine if it should skip.

Each environment should still accept an explicit true or false that would enable/disable it for all regions in all accounts.

Each environment should be able to accept an accountId only or a region only.

ITProKyle:
## Why This Is Needed

resolves #333 

### Added

- option for how to handle `environments` being defined but the current environment not in the definition

### Changed

- log format of skip information no longer includes extra characters and is now prefixed by the module name

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ integration_tests/test_cfngin/tests/[0-9][0-9]_*.yaml
 
 # Programatically generated doc pages
 docs/source/apidocs
+
+# vscode settings
+.vscode/bookmarks.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `-no-color`/`--no-color` option automatically added to cdk, npm, sls, and tf commands
   - looks at `RUNWAY_COLORIZE` env var for an explicit enable/disable
   - if not set, checks `sys.stdout.isatty()` to determine if option should be provided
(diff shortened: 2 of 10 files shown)


ITProKyle:
macOS failure is likely on GitHub's end, there arn't any logs that that show a fault.

ITProKyle:
not sure where _static_ came from... looks like it only got into the docs/docstring though. i'll swap it out.

PR #346, opened 2020-05-28
add message about false alarm pip errors & pip --no-color

ITProKyle:
## Why This Is Needed

resolves #345 

### Added

- notice about *false alarm* compatibility errors when using pip with the aws_lambda hook

### Changed

- aws_lambda hook will no longer use colorized pip output so its *false alarm* compatibility errors are less menacing

PR #350, opened 2020-05-29
disable pip red error messages; add pip warning

### LINKED ISSUE #345: Static site Auth@Edge reports a red "ERROR" during lambda builds
Deploying an auth@edge-enabled static site (at least on my system -- Runway 1.8 in pipenv on Ubuntu 20.04) generates the following output (among a bunch of other lines):

```
INFO:runway.cfngin.hooks.aws_lambda:lambda: using custom bucket: mynamespace-frontend-dependencie-artifacts-18wze3vbi33ga
INFO:runway.cfngin.hooks.aws_lambda:lambda: using requirements.txt for dependencies
Collecting ecdsa==0.15
  Using cached ecdsa-0.15-py2.py3-none-any.whl (100 kB)
Collecting pyasn1==0.4.8
  Using cached pyasn1-0.4.8-py2.py3-none-any.whl (77 kB)
Collecting pyjwt==1.7.1
  Using cached PyJWT-1.7.1-py2.py3-none-any.whl (18 kB)
Collecting python-jose==3.1.0
  Using cached python_jose-3.1.0-py2.py3-none-any.whl (26 kB)
Collecting rsa==4.0
  Using cached rsa-4.0-py2.py3-none-any.whl (38 kB)
Collecting six==1.14.0
  Using cached six-1.14.0-py2.py3-none-any.whl (10 kB)
ERROR: runway 1.8.0 requires zipp~=1.0.0, which is not installed.
ERROR: awscli 1.18.62 has requirement rsa<=3.5.0,>=3.1.2, but you'll have rsa 4.0 which is incompatible.
Installing collected packages: six, ecdsa, pyasn1, pyjwt, rsa, python-jose
Successfully installed ecdsa-0.15 pyasn1-0.4.8 pyjwt-1.7.1 python-jose-3.1.0 rsa-4.0 six-1.14.0
INFO:runway.cfngin.hooks.aws_lambda:lambda: object lambda-check_auth-be24ce1bd52bfe0b7131c22df54e9778.zip already exists, not uploading
```

The 2 ERROR lines from pip are colored red. The RSA conflict [is due to an awscli issue](https://github.com/aws/aws-cli/issues/3660); I'm not sure what's causing the zipp issue but I don't think it's important.

troyready:
Closes #345
Supersedes #346

### DIFF
--- a/integration_tests/test_cfngin/tests/16_aws_lambda_hook.py
+++ b/integration_tests/test_cfngin/tests/16_aws_lambda_hook.py
@@ -1,15 +1,17 @@
 """Test AWS Lambda hook."""
 # flake8: noqa
 # pylint: disable=invalid-name
-from os import path
-from os.path import basename
+import os
+import pty
+import tempfile
 
 import boto3
 from send2trash import send2trash
 
 from integration_tests.test_cfngin.test_cfngin import Cfngin
+from runway.util import environ
 
-FILE_BASENAME = '.'.join(basename(__file__).split('.')[:-1])
+FILE_BASENAME = '.'.join(os.path.basename(__file__).split('.')[:-1])
 
 
 class TestAwsLambda(Cfngin):
@@ -33,10 +35,27 @@ class TestAwsLambda(Cfngin):
         return resp['StatusCode']
 
     def _build(self):
-        """Execute and assert initial build."""
-        self.set_environment('dev')
-        code, _stdout, _stderr = self.runway_cmd('deploy')
-        assert code == 0, 'exit code should be zero'
+        """Execute and assert initial build.
+
+        Explicitly spawning with a tty here to ensure output
+        (e.g. from pip) includes color
+
+        """
+        with environ({**os.environ, **{'DEPLOY_ENVIRONMENT': 'dev'}}):
+            with tempfile.TemporaryFile() as pty_output:
+                def read_pty(fds):
+                    """Append tty output to file descriptor."""
+                    data = os.read(fds, 1024)
+                    pty_output.write(data)
+                    return data
+
+                spawn_result = pty.spawn(['runway', 'deploy'], read_pty)
+                assert spawn_result == 0, 'exit code should be zero'
+                pty_output.seek(0)
+                combined_output = pty_output.read().decode()
+                assert '\x1b[31mERROR: ' not in combined_output, (
+                    'no red ERROR should be present'
+                )
 
     def run(self):
         """Run test."""
@@ -46,7 +65,8 @@ class TestAwsLambda(Cfngin):
         client = boto3.client('lambda', region_name=self.region)
 
         functions = ['dockerizepip-integrationtest',
-                     'nondockerizepip-integrationtest']
+                     'nondockerizepip-integrationtest',
+                     'authatedge-integrationtest']
         for func_name in functions:
             assert self.invoke_lambda(client, func_name) == 200, \
                 f'{self.TEST_NAME}: Execution of lambda {func_name} failed'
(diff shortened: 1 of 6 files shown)


runway/cfngin/hooks/aws_lambda.py
```diff
@@ -464,6 +464,29 @@ def dockerized_pip(work_dir, client=None, runtime=None, docker_file=None,
...
+             'import pip;'
+             'print(pip.__version__)']
+        )
+        if sys.version_info[0] > 2 and isinstance(pip_version_string, bytes):
+            pip_version_string = pip_version_string.decode()
+        if int(pip_version_string.split('.')[0]) > 10:
+            return True
+    except (AttributeError, ValueError, subprocess.CalledProcessError):
+        LOGGER.debug('Error checking pip version; assuming it to be pre-v10')
+    return False
```
ITProKyle, line 487: There are better and more accurate options available to get a python packages version list `pkg_resources` or `importlib.metadata`.

Using `importlib`, you can add an import block like this

```python
if sys.version_info < (3, 8):
    # importlib.metadata is standard lib for python>=3.8, use backport
    from importlib_metadata import version, PackageNotFoundError  # pylint: disable=E
else:
    from importlib.metadata import version, PackageNotFoundError  # pylint: disable=E
```

and reduce the logic needed to determine support to this

```suggestion
    try:
        return int(version('pip').split('.')[0]) >= 10
    except PackageNotFoundError:
        LOGGER.debug('Error checking pip version; assuming it to be pre-v10')
        return False
```

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -464,6 +464,29 @@ def dockerized_pip(work_dir, client=None, runtime=None, docker_file=None,
         LOGGER.info('lambda.docker: %s', log.decode().strip())
 
 
+def _pip_has_no_color_option(python_path):
```
ITProKyle, line 467: This probably does not need to be a protected function and it should have test written for it.

runway/cfngin/hooks/aws_lambda.py
troyready: The supplied `python_path` here comes from the `_zip_package` function, where it may be different than what's currently running this code. I don't see a safer path than using subprocess in the same way that _zip_package is, but I'm open to suggestions.

runway/cfngin/hooks/aws_lambda.py
troyready: I was torn on whether to make it open. I decided not to because I think we're likely to toss it entirely in the relatively near future, and its so dedicated to this purpose that I don't want any external use of the function to impede us getting rid of it.

I'll see about testing it

runway/cfngin/hooks/aws_lambda.py
ITProKyle: right... I  guess we did provide the option to specify a custom python path.

>  I think we're likely to toss it entirely in the relatively near future

if we are going to toss this in the near future, i am fine leaving this as is

runway/cfngin/hooks/aws_lambda.py
ITProKyle: >  I think we're likely to toss it entirely in the relatively near future

if this is the case, we should have a timeline for removal added noted somewhere.

honestly, not sure who could put up with >2 years of pip telling you to update it each time something is installed to still be using <v10.

troyready: > honestly, not sure who could put up with >2 years of pip telling you to update it each time something is installed to still be using <v10.
this.jpg

troyready:
I've taken a first pass at an integration test for the issue. I didn't make the tty invocation stuff more general (not sure yet how much we'd want to use it), but I'd be open to doing it now or later.

integration_tests/test_cfngin/tests/16_aws_lambda_hook.py
```diff
@@ -1,61 +1,81 @@
-"""Test AWS Lambda hook."""
```
troyready, line 1: This file is not nearly as changed as the git diff is implying here. The main change is the `_build` method.

Issue #352, opened 2020-06-02
Angular gen-sample template is out of date


troyready:
`runway gen-sample static-angular` generates an Angular 8 app, instead of a current v9 app.

Instead of just updating it for the latest version, it may be more future-proof to remove the code and either:

1) Print instructions for the user to generate the project themselves (e.g. `npx ng new APPNAMEHERE --skip-git --strict --directory frontend`)
or
2) Prompt the user for info and run the command ^ on their behalf

PR #353, opened 2020-06-02
sls-tsc template: migrate from tslint to typescript-eslint

### LINKED ISSUE #261: Typescript samples should not include tslint
E.g. `gen-sample sls-tsc`

TSLint has been [deprecated in favor of eslint](https://github.com/palantir/tslint/issues/4534). Any tslint examples/use should be converted to eslint directly.

troyready:
Fixes #261 (the remaining reference to tslint can be addressed in #352)

Also bumped up some package versions in the sample to address deprecation/security warnings.


troyready:
Not sure why the macOS GitHub action was cancelled (I certainly didn't do it); possibly a transient GH actions bug?

PR #368, opened 2020-06-24
staticsite extra_files

paul-duffy:
## Summary

Add a new module option to allow users to specify dynamic or static files that should be uploaded to S3 after a build. These files would typically contain environment specific configurations that can help configure the static site at runtime.

## Why This Is Needed

Static sites typically need to be compiled per environment and don't get the benefit of artifact promotion through the environments: dev -> test -> prod. This creates a standard way of providing environment specific configurations to static static sites and allows the artifact promotion.

### Added

- `extra_files` staticsite option
- documentation an examples for staticsite `extra_files`
- `hooks.staticsite.upload_staticsite.sync` handling for `extra_files`


runway/hooks/staticsite/upload_staticsite_env.py
```diff
@@ -0,0 +1,70 @@
...
+                content = json.dumps(content)
+                content_type = 'application/json'
+            elif ext in ['.yml', '.yaml']:
+                content = yaml.dump(content)
+                content_type = 'text/yaml'
+            else:
+                LOGGER.debug('type: %s', type(content))
+                content = str(content)
+
+            if content is not None:
```
ITProKyle, line 53: would this be better as `if content:` to prevent uploading empty files or are empty files ok?

runway/hooks/staticsite/upload_staticsite_env.py
```diff
@@ -0,0 +1,70 @@
...
+
+            if ext == '.json':
+                content = json.dumps(content)
+                content_type = 'application/json'
+            elif ext in ['.yml', '.yaml']:
+                content = yaml.dump(content)
+                content_type = 'text/yaml'
+            else:
+                LOGGER.debug('type: %s', type(content))
+                content = str(content)
```
ITProKyle, line 51: i would almost make content_type need to be explicit, check the type before dumping, and ensure we were able to correctly convert it to a string before continuing. i would also be fine with a combination of autodetecting the content_type and explicitly providing it.

reasoning: there could a file that is json or yaml formatted that does not end with an expected extension. if we skip the last string check and instead just `str()` whatever is there without proper handling of the conversion, it could end up being a useless mess of a file.

```suggestion
        if env_file.get('content'):
            content = env_file['content']
            content_type = env_file.get('content_type', 'text/plain')

            if isinstance(content, (dict, list)):
                if content_type == 'application/json':
                    content = json.dumps(content)
                elif content_type == 'text/yaml':
                    content = yaml.safe_dump(content)
                else:
                    raise ValueError('"content_type" must be provided if '
                                     '"content" is not a string')
            if not isinstance(content, str):
                raise TypeError('unsupported content type: %s' % type(content))
```

runway/hooks/staticsite/upload_staticsite_env.py
```diff
@@ -0,0 +1,70 @@
...
+
+            if content is not None:
+                LOGGER.info(
+                    'Uploading environment file as %s with content: %s', filename, content)
```
ITProKyle, line 55: i would reserve logging the content to a `debug` message. it could be a lot of spam for the info level.

runway/hooks/staticsite/upload_staticsite_env.py
```diff
@@ -0,0 +1,70 @@
+"""CFNgin hook for uploading environment files to a static website S3 bucket."""
```
ITProKyle, line 1: does this need its own file? it seems like it could be a new function in `upload_staticsite.py`.

also, shouldn't this be performing some kind of hash calculation/cache invalidation so that the cloudfront distribution is updated when these files change?

add documentation for this new hook to [this page](https://docs.onica.com/projects/runway/en/release/cfngin/hooks.html).

docs/Pipfile
```diff
@@ -8,5 +8,6 @@ Sphinx = "~=2.3"
 sphinx_rtd_theme = "===0.4.2"
 # just for http.server py2/py3 syntax consistency in manual doc testing
 future = "~= 0.16"
-runway = {editable = true, path = "./.."}
+runway = {editable = true,path = "./.."}
 sphinxcontrib-apidoc = "*"
+jsx-lexer = "*"
```
ITProKyle, line 13: `docs/requirements.txt` needs to be updated with this change.

run `pipenv lock --requirements > requirements.txt` from `docs/`. you will then need to manually remove the entry for `runway`. iirc it gets added to the top of the file as a relative path (e.g. `../.`)

docs/source/staticsite/configuration.rst
```diff
@@ -61,6 +61,33 @@ Options
             exclusions:
               - foo/*
 
+**environment_files (Optional[List[Dict[str, str]]])**
```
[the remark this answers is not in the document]
troyready: I agree with the name overlap concern.

I can't think of a perfect alternative yet, but to try to spur some discussion, one route would be to make it more explicit. e.g. `staticsite_environment_files`

Or maybe `staticsite_config_files`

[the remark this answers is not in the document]
ITProKyle: a few other potential names/name elements that are descriptive yet avoid some of our currently overused words (environment and config).

- `dynamic_files`
- `post_deploy_dynamic_files`
- `extra_files`

runway/hooks/staticsite/upload_staticsite_env.py
paul-duffy: I added in support to invalidate the cache. This doesn't do any kind of hash checking though. Basically, any time a file is uploaded is the trigger for when the cache is invalidated. I think we can optimize invalidation later in a different PR.

runway/hooks/staticsite/upload_staticsite.py
```diff
@@ -201,3 +223,174 @@ def prune_archives(context, session):
...
+            instance.
+        bucket (str): The static site bucket name.
+        extra_files (List[Dict[str, str]]): List of files and file content that should be
+            uploaded.
+
+    """
+    LOGGER.debug('bucket: %s', bucket)
+    LOGGER.debug('extra_files: %s', json.dumps(extra_files))
+
+    if len(extra_files) == 0:
```
ITProKyle, line 333: should be this

```suggestion
    if not extra_files:
```

runway/hooks/staticsite/upload_staticsite.py
```diff
@@ -42,37 +46,55 @@ def sync(context, provider, **kwargs):
...
+    invalidate_cache = False
+
+    extra_files = sync_extra_files(
+        context,
+        bucket_name,
+        kwargs.get('extra_files', []),
+        hash_tracking_parameter=build_context.get('hash_tracking_parameter')
+    )
+
+    if len(extra_files) > 0:
```
ITProKyle, line 63: should be this

```suggestion
    if not extra_files:
```

runway/hooks/staticsite/upload_staticsite.py
```diff
@@ -42,37 +46,55 @@ def sync(context, provider, **kwargs):
...
+        invalidate_cache = True
 
         LOGGER.info("staticsite: sync " "complete")
 
         update_ssm_hash(context, session)
 
+    if kwargs.get('cf_disabled', False):
+        display_static_website_url(kwargs.get('website_url'), provider, context)
+
+    if invalidate_cache:
```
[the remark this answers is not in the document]
paul-duffy: Good catch, that should be `elif`.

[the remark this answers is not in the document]
paul-duffy: Also, invalidate_cache is initialized to False. There are 2 independent scenarios when the cache should be cleared:

1. Any of the extra files change
2. Any of the static site files change

If any of those are true, then the cache will be invalidated. This makes sure that if both are True, we don't have to calls to invalidate the cache.

runway/hooks/staticsite/upload_staticsite.py
```diff
@@ -201,3 +223,174 @@ def prune_archives(context, session):
...
+            raise TypeError('unsupported content: %s' % type(content))
+
+    return content
+
+
+def calculate_hash_of_extra_files(extra_files):
+    """."""
+    file_hash = hashlib.md5()
+
+    for extra_file in sorted(extra_files, key=lambda extra_file: extra_file['name']):
```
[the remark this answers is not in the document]
paul-duffy: This was basically taken  from `calculate_hash_of_files` in `hooks/staticsite/util.py` That method also sorts by filename. I don't think that someone changing the order of the `extra_files` list while keeping everything else the same should make the hash be different. Sorting by name will ensure that the hash will be the same in that scenario.

runway/hooks/staticsite/upload_staticsite.py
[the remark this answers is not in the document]
ITProKyle: i think i follow how `invalidate_cache` is being determined now. couldn't you use `elif (extra_files or not build_context['deploy_is_current']):` instead of needing the extra variable so that all the decision logic is in the statement.

PR #372, opened 2020-06-29
allow runway tests to exit 0 without signaling failure

troyready:
This will check the exit code of tests so any that have called sys.exit(0) (e.g. yamllint) won't trigger Runway to count it as failed.


runway/commands/runway/test.py
```diff
@@ -78,6 +78,8 @@ def execute(self):
                 # tool it is wrapping.
                 if not isinstance(err, SystemExit):
                     traceback.print_exc()
+                elif err.code == 0:
+                    next  # Tests calling sys.exit(0) don't indicate failure
```
ITProKyle, line 82: I think you are looking for `continue` here

```suggestion
                    continue
```

PR #374, opened 2020-07-01
update macOS python setup docs to default to zsh

troyready:
I also considered making the shell code smarter here (e.g. a condition that checks `$SHELL`) but I wasn't sure if that'd be overkill.

### DIFF
--- a/docs/source/python_setup.rst
+++ b/docs/source/python_setup.rst
@@ -12,11 +12,12 @@ your user will be able to install additional Python packages via
 
 1. Ensure you have a working python environment:
     - On macOS:
+        - Note: Substitute ``.bash_profile`` in place of ``.zsh`` below if using Bash (e.g. macOS pre-Catalina)
         - Add local python bin directories to $PATH::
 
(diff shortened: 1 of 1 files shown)


docs/source/python_setup.rst
```diff
@@ -12,11 +12,12 @@ your user will be able to install additional Python packages via
 
 1. Ensure you have a working python environment:
     - On macOS:
+        - Note: Substitute ``.bash_profile`` in place of ``.zsh`` below if using Bash (e.g. macOS pre-Catalina)
         - Add local python bin directories to $PATH::
 
-            if ! grep '\.local\/bin' ~/.bash_profile > /dev/null 2>&1 || ! grep 'Library\/Python\/2\.7\/bin' ~/.bash_profile > /dev/null 2>&1 ; then echo 'export PATH="$HOME/Library/Python/2.7/bin:$HOME/.local/bin:$PATH"' >> ~/.bash_profile; fi
+            if ! grep '\.local\/bin' ~/.zsh > /dev/null 2>&1 || ! grep 'Library\/Python\/2\.7\/bin' ~/.zsh > /dev/null 2>&1 ; then echo 'export PATH="$HOME/Library/Python/2.7/bin:$HOME/.local/bin:$PATH"' >> ~/.zsh; fi
```
ITProKyle, line 18: I believe this should be `~/.zshrc` not `~/.zsh`.

Issue #300, opened 2020-05-18
[REQUEST] refactor logging


ITProKyle:
# Acceptance Criteria

- [x] colorized logs
	- [x] user configurable
- [x] add more log levels for more variation in colors
	- [x] verbose
	- [x] notice
	- [x] success
- [x] standardize formatting across runway
	- [x] user configurable
- [x] add more debug/verbose logs throughout runway for improved troubleshooting
- [x] review and update current info/warning messages to see if any should be moved to one of the new levels to remove some spam from the info level
- [x] display level should be configurable by CLI option or env var (complete in tandem with #299 )
- [x] 100% test coverage of everything touched

PR #375, opened 2020-07-01
refactor cli to improve ux & separate logic

### LINKED ISSUE #299: [REQUEST] refactor command line interface
This can be done with use any library or combination of libraries but I have a preference for [click](https://click.palletsprojects.com/en/7.x/) currently.

# Acceptance Criteria

- [x] each command should have their own `--help` output with detailed information pertaining to the command instead of the single page currently provided by docopt
- [x] simplify prompts & confirmations by using methods built into the library
- [x] tab completion
- [x] all environment variables should also be CLI options (advanced options can be hidden by default)
- [x] 100% test coverage of everything touched

ITProKyle:
## Summary

Improve the user experience and testability of Runway by switching CLI libraries from docopt to click. Runway's core logic has also been separated from the CLI to allow for easier testing via direct API calls.

## Why This Is Needed

resolves #299 

# What Changed

## Added

- `--ci` cli option to enable non-interactive mode from the cli rather than requiring an environment variable to be set before hand (which is still supported)
- `-e, --deploy-environment` cli option to pass a deploy environment name from the cli rather than requiring an environment variable to be set before hand (which is still supported)
- `docs` command to open the doc site using the default browser

## Changed

- using `-h, --help` from each command will now display a command specific help page rather than the previous "one page to rule them all"
- some log messages were changed to be more consistent
- some log messages were changed from _info_ to _debug_ to remove some repetitive or long messages from the info level
- many log messages now use a logger specific to their file instead of logging everything to the `runway` parent logger for more detailed information
- `runway destroy` initial confirmation prompt is now printed rather than logged and will display as red if the terminal supports it
- `runway plan/taxi` will log that processing will happen in parallel for deploy/destroy but will not run in parallel itself to prevent intermingled outputs

## Removed

- removed some repetitive log messages
- removed the "one page to rule them all" `-h, --help` help page


ITProKyle:
I am currently running through the tests in the `integration_tests/` that are run using CodeBuild upon release to ensure they are still functional.

There is an issue with the `integration_tests/test_cfngin/tests/16_aws_lambda_hook.py` test. As is, it does not cd into the correct directory for the test to run successfully. I patched this locally only to encounter another issue - it hangs at the end of running `runway deploy` (this happened without the cd as well) as the process spawed by pty is not ending. The test does succeed with the my patch that cd's into the test directory but requires manually ending the process via keyboard input for the test to complete.

ITProKyle:
starting to work on #300 as a follow up to this PR. while the code in this PR is required for it, release of the feature could be held back to land both the CLI and log changes in the same release.

PR #376, opened 2020-07-02
add additional terraform backend init arguments

troyready:
Now key / prefix can be managed in runway.yml, in addition to the parameters for assuming a role.


troyready:
I don't think most of these will be used much, but it's more intuitive for the options to be available.

Issue #379, opened 2020-07-15
[REQUEST] Terraform Enterprise backend support


aoggz:
**Is your feature request related to a problem? Please describe.**
It appears that runway only supports terraform backends of type s3, currently. My organization has acquired Terraform Enterprise, and we are working on migrating our Terraform state storage from S3 to Terraform Enterprise. Since we heavily use runway, this lack of support is causing us trouble.

**Describe the solution you'd like**
Update the `terraform_backend_config` property (and the backend config file feature) to support [remote backend properties](https://www.terraform.io/docs/backends/types/remote.html#basic-configuration).

**Describe alternatives you've considered**
A few things I've tried to get around this limitation:

- Passing a `backend-config` arg to terraform ([despite the warning in the docs](https://docs.onica.com/projects/runway/en/release/terraform/advanced_features.html#specifying-terraform-cli-arguments-options))
- Adding a `backend[-*].tfvars` file with the best guesses at appropriate variable names
- Running `terraform init` for terraform modules before running `runway deploy`

aoggz:
So it looked like your suggestion to use `backend-*.tfvars` worked for getting the backend parameters correct, but runway failed shortly thereafter:
```sh
[runway] infrastructure:workspace currently set to default; switching to my-awesome-workspace...
workspaces not supported
```
Terraform enterprise/cloud workspace do not allow you to switch workspaces: [only `default` is permitted/relevant in the given cloud workspace](https://www.terraform.io/docs/backends/types/remote.html#workspaces).

That said, there are other issues that I'm realizing. We've leaned heavily on using runway's variable & interpolation in favor of using `*.tfvars` or other module-specific variable syntax. Terraform cloud/enterprise does not support specifying those variables via CLI. They must be specified via `*.auto.tfvars` files or specified in the cloud/enterprise workspace manually (or at least before you try to use the given workspace for a `plan`/`apply`).

While I was able to overcome the reported shortcoming (thank you!), it looks like there's more to it than I initially realized. At least as far as that output I shared above goes.

The ideal use case for my team is:

- The `terraform_backend_config` config option is updated to support the new types required for the `remote` backend.
- The backend logic is updated to not switch workspaces once `terraform init` is run for backends of type `remote`
- A `*.auto.tfvars` file is created based on the `parameters` I provide in the `runway.yml`/`runway.variables.yml` files, where `*` is the runway environment/terraform cloud workspace name.

This is pretty opinionated, I know.

[...]

PR #390, opened 2020-07-22
update env managers to use pathlib & have useful attributes

ITProKyle:
## Why This Is Needed

pathlib provides easier manipulation of the file system.

env managers now has attributes for getting the binary path and current version number instead of needing to parse the return value of `.install()` to extract the information since it has already been parsed by the env manager then joined for the return. I need the current version of Terraform selected for another feature being worked on. Updating this to be more useful made more sense than parsing the return value again.

### Added

- added attributes to the `EnvManager` base class for getting the binary path as a path object and the current version number

### Changed

- `EnvManager` and it's subclasses now use `pathlib` for file/path manipulation instead of `os`
- moved function logic from `runway.env_mgr.__init__.py` into the `EnvManager` base class for ease of use

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- env managers now use pathlib
+- env managers have some new attributes and methods for handling envs (relocated from functions)
 
 ## [1.10.1] - 2020-07-20
 ### Fixed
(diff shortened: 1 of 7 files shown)


ITProKyle:
> I need the current version of Terraform selected for another feature being worked on.

I think this is less important as I have readjusted my approach. It was going to be used to identify if the Terraform version supports `*.auto.tfvars` but, dynamically generating one of these files in the module directory would break parallel deployment as it is automatically loaded by Terraform. That is if there arn't already other complications with parallel deployments.

ITProKyle:
After looking into supporting the remote backend (#379) a bit further and running tests, use of a `*.auto.tfvars` is required to pass any variables so its use is going to be unavoidable; rendering parallel execution unusable if the values need to differ between environments/regions.

ITProKyle:
superseded by #392

PR #396, opened 2020-07-30
move prod release to it's own workflow triggered on github release

### LINKED ISSUE #327: [TODO] change publishing to pypi, npm, s3, etc to be trigger on release instead of tag push
This will require a separate workflow that is triggered when a release is published on GitHub. The release flow will change to require someone to publish the release from GitHub before anything is published to external sources.

Example workflow

```yaml
name: Publish Release

on:
  release:
    types:
      - published

jobs:
  ...
```

~This works when releases are published from `master`, will need to ensure it will work from the `release` branch.~ Releases are tied to tags so this should not matter.

ITProKyle:
## Summary

resolves #327

Final artifact building and publishing has been moved out of the `CI/CD` workflow and moved into the `Publish Release` workflow.

Previously, it relied on conditions to determine if the steps should be run or skipped and those that were skipped still always appears in the checks of each PR. This removes that.

Moving the trigger from `tag` to `release.published` better restricts when these steps are run and forces the use of GitHub releases which notifies users of new versions and can be used to track change instead of keeping a changelog.

### Changed

- copied build steps to a new workflow
- moved production publishing steps from `CI/CD` workflow to a new workflow

### DIFF
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -1,10 +1,9 @@
-name: CI/CD
+name: Publish Release
 
 on:
-  pull_request:  # any pull request
-  push:
-    branches:
-      - master
+  release:
+    types:
+      - published
 
 env:
   AWS_DEFAULT_REGION: us-east-1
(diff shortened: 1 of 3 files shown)


ITProKyle:
@troyready - its not something we have to move to immediately but it does remove manual steps of writing down what was done in both the CHANGELOG and PR. Arguably, the PR will contain more data more easily as it can be linked to an issue and have conversations attached as to why things were done.

As for being grep-able, the body of each release can be collected with `curl -s https://api.github.com/repos/onicagroup/runway/releases`. It's not a standalone file but still just as effective.

PR #420, opened 2020-08-19
Update other_ways_to_use.rst

dgreeninger:
Removing mention of the public docker image, as it is out of date and not actively maintained. 

## Summary

There are no plans to upgrade the docker image with the latest runway versions, so we should not include it in the docs.

## Why This Is Needed

Improves overall quality of docs and does not encourage users to use an old version of runway.

## What Changed

Removed mention of publicly available container.

### DIFF
--- a/docs/source/quickstart/other_ways_to_use.rst
+++ b/docs/source/quickstart/other_ways_to_use.rst
@@ -29,7 +29,7 @@ Docker users can build their own Docker image to run a local Runway
 container or modify this `Dockerfile`_ to build a Runway image to suit specific
 needs.
 
-We have also provide a pre-build Docker image on `Docker Hub`_ that can be
+We have also provide a pre-built Docker image on `Docker Hub`_ that can be
 used with the following command.
 
 .. code-block:: shell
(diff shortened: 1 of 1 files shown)


troyready:
Thanks @dgreeninger !

For some additional context, in case anyone comes across this PR looking for it, here's a brief message I drafted previously on why we're not maintaining public docker images at this time:

The issue is that outside of CloudFormation, Runway works with a variety of supporting tools (kubectl, node/npm, terraform, etc) and we've never wanted to bite off supporting the matrix of images - e.g. runway:1.x-tf0.12.x-node12.x, runway:1.x-tf0.11.x-node10.x, etc. Even if we did that, it would still likely steer away from our best practices in areas like Terraform management, where it'd be better for clients to have custom images with providers cached.

It was a big deal before since asking users to manage a python installation is non-trivial. But given that installation is as simple as `curl -L oni.ca/runway/latest/linux -o /usr/local/bin/runway && chmod +x /usr/local/bin/runway`, there's no barrier to entry to use it in an existing image or bake it into a custom one (or copying that binary to S3 to keep it archived in the deployment account or whatever)

PR #435, opened 2020-08-28
drop support for python < 3.6

### LINKED ISSUE #294: [REQUEST] drop python 3.5 support
Like python2, support for 3.5 is increasingly being left behind by runway's dependancies. The must recent being ``zipp``, a dependency of ``importlib-metadata``.

We should raise or new minimum required version to at least 3.6 which is generally more supported.

We have already stopped testing for 3.5, the next step is to modify the required python version in `setup.py`.

### LINKED ISSUE #295: [REQUEST] drop python 2 support
- ETA December 31, 2020
- must be accompanied by a major version bump

ITProKyle:
## Summary

Remove support for Python < 3.6.

## Why This Is Needed

- resolves #295 
- resolves #294 

### Changed

- the required python version is now `>=3.6`
- dependency markers and pins have been updated to no longer account for python 2
- `TODO`s referencing python 2 have been resolved

### Removed

- tests are no longer run for 2.7
- zipp is no longer a dependency - it was only added to pin the version for python 2.7/3.5 support
- six is no longer a dependency and is no longer used
- pathlib2 is no longer used
- `future`/`__future__`/`past` is no longer used


ITProKyle:
The current build errors are caused by the version of pyinstaller we are stuck on. It does not work with modern pytest.

Possible solutions:

1. Wait for a usable new version of pyinstaller.
2. Change the build process to no install dev dependencies which should probably be done either way.

ITProKyle:
I am going to merge this for now. any alterations can be handled in future PRs. We will need to create separate issues for resolving any incoming changes that will impact this by including handling for python 2.

PR #442, opened 2020-09-04
fix `YamlDirTestGenerator` python 3 compatibility

twitty-onica:
## Summary

Use of `dict.iteritems()` was removed in Python 3. This change updates the reference to use `dict.items()` instead.

## Why This Is Needed

Class `YamlDirTestGenerator` is broken in Python 3 due to legacy `dict.iteritems()` use.

### Fixed

- fixed Python 3 compatibility issue in `YamlDirTestGenerator` when iterating a `dict`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- fixed a python 3 compatibility issue in `runway.cfngin.blueprints.testutil.YamlDirTestGenerator`
 
 ## [1.12.0] - 2020-09-11
 ### Changed
--- a/runway/cfngin/blueprints/testutil.py
+++ b/runway/cfngin/blueprints/testutil.py
@@ -150,7 +150,7 @@ class YamlDirTestGenerator(object):
 
                 configvars = self.stack.variables or {}
                 variables = [
-                    Variable(k, v, "cfngin") for k, v in configvars.iteritems()
+                    Variable(k, v, "cfngin") for k, v in configvars.items()
                 ]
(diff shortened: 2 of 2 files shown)


ITProKyle (changes requested):
Please fix the linting error.

twitty-onica:
I'm not sure how to appease the CI/CD's linter. I ran `make fix-black` and it did all these formatting changes. Now when I run `make lint-black` it shows
```
(runway) ➜  runway git:(bugfix/iteritems) make lint-black
Running black... If this fails, run 'make fix-black' to resolve.
Loading .env environment variables…
All done! ✨ 🍰 ✨
435 files would be left unchanged.
```

Am I missing a step somewhere?

ITProKyle:
@twitty-onica I think you may be using the wrong version of black. There were some changes made in the newest release of black that made its formatting semi-unstable causing a lot of unnecessary change which is why we are not using it.

Please revert the most recent commit to remove the unnecessary changes and ensure you are setting up your development environment correctly. If you are unsure how to, I suggest reviewing the [docs](https://docs.onica.com/projects/runway/en/release/developers/getting_started.html#development-environment).

twitty-onica:
Thanks Kyle, I'll take a look when I get a chance.

twitty-onica:
FYI, the latest version of pipenv has a bug that prevents `make sync_all` from working. I had to roll back to an older version.
https://github.com/pypa/pipenv/issues/4326

twitty-onica:
In order to run `make fix-black` I need to have python 3.6 installed. This is not setup on my machine (running python 3.8.4) and I unfortunately, don't have time right now to figure out how to install different versions of python for different projects.
I'll try to get to it when I have some spare time.

ITProKyle (approved):
Thanks for this. The remaining failure is not due to your change but looks to be another dependency + python 2 conflict. I'll track it down when I get a chance and update this branch with the fix so it can pass before merging (unless Troy merges it before then).

PR #444, opened 2020-09-08
fix cfngin kms lookup on py3

troyready:
This should avoid returning a binary value unexpectedly.

### DIFF
--- a/runway/cfngin/lookups/handlers/kms.py
+++ b/runway/cfngin/lookups/handlers/kms.py
@@ -1,6 +1,7 @@
 """AWS KMS lookup."""
 # pylint: disable=arguments-differ,unused-argument
 import codecs
+import sys
 
 from runway.lookups.handlers.base import LookupHandler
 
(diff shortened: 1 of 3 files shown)


runway/cfngin/lookups/handlers/kms.py
```diff
@@ -73,4 +74,7 @@ def handle(cls, value, context=None, provider=None, **kwargs):
         decoded = codecs.decode(value, "base64")
 
         # decrypt and return the plain text raw value.
-        return kms.decrypt(CiphertextBlob=decoded)["Plaintext"]
+        decrypted = kms.decrypt(CiphertextBlob=decoded)["Plaintext"]
+        if sys.version_info[0] > 2:  # TODO remove after droping python 2
+            return decrypted.decode("UTF-8")
+        return decrypted
```
ITProKyle, line 80: could do something like this to remove the version check and instead rely on the data's type since it should always return a string.

```suggestion
        return decrypted.decode() if isinstance(decrypted, bytes) else decrypted
```

`UTF-8` can also be removed as its the default.

ITProKyle (changes requested):
looks like the test needs to be updated to `.decode()` as well.

PR #448, opened 2020-09-14
fix issue causing static sites to be inaccessible when not us-east-1

### LINKED ISSUE #445: [BUG] creating a static website in a region outside of us-east-1 doesn't work.
**Describe the bug**
Creating a static website in a region outside of us-east-1 doesn't work.

**To Reproduce**
1. git clone https://github.com/burib/runway-static-website-bug
2. npm install
3. npm run deploy
4. check both cloudfront distributions
5. wonder why is us-east-1 but the rest doesn't work. ( my hunch is that it's related to OAI creation )

**Expected behavior**
Static website creation with cloudfront enabled works in all regions.

**Desktop (please complete the following information):**
 - OS: osx
 - Version 10.15.6

**Additional context**
code to reproduce the bug is here: https://github.com/burib/runway-static-website-bug

ITProKyle:
## Why This Is Needed

resolves #445 

[image](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html)

According to the [AWS docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html) screenshot above, our current implementation of configuring the S3 bucket for hosting a static website conflicts with the use of an OAI to grant access.

By removing the `WebsiteConfiguration`, we can use the OAI correctly.

To continue using `WebsiteConfiguration` with CloudFront, we would need to make the bucket private and/or use custom headers to restrict access - configuring the Bucket as a custom origin rather than an S3 origin.

### Changed

- the static site S3 Bucket is now only configured for static website hosting if not using CloudFront

### Fixed

- fixed an issue causing static sites to be inaccessible when deployed to regions other than us-east-1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Fixed
 - fixed a python 3 compatibility issue in `runway.cfngin.blueprints.testutil.YamlDirTestGenerator`
-
-### Fixed
 - fixed an issue causing static sites to be inaccessible when deployed to regions other than us-east-1
 
 ## [1.12.0] - 2020-09-11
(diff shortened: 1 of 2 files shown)


troyready (approved):
:heart:

PR #451, opened 2020-09-15
static site CF origin will now use the s3 regional endpoint to avoid temporary redirect for new deployments

### LINKED ISSUE #445: [BUG] creating a static website in a region outside of us-east-1 doesn't work.

ITProKyle:
## Why This Is Needed

resolves #445 ([second problem raised in the issue](https://github.com/onicagroup/runway/issues/445#issuecomment-692826339))

### Changed

- static site CloudFront distribution now uses regional endpoints for the S3 bucket
	- us-east-1 still uses the global endpoint 

### Fixed

- fixed an issues causing static sites in regions other than us-east-1 to redirect to the s3 object until CloudFront was able to use the global endpoint of the bucket

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- fixed an issues causing static sites in regions other than us-east-1 to redirect to the s3 object until CloudFront was able to use the global endpoint of the bucket
 
 ## [1.13.0] - 2020-09-14
 ### Added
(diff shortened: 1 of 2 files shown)


troyready (approved):
:100:

burib:
👏 👏  could you guys bump the version number to 1.13.1 so I can test it ?

ITProKyle:
@burib - I don't have an ETA on the official release but this has been published in the most recent dev release on npm & the test instance of pypi. To install it from npm it would be `npm i @onica/runway@1.13.1-dev3`.

PR #455, opened 2020-09-16
use pydantic for config validation

### LINKED ISSUE #431: [TODO] investigate using pydantic for config file validation & implement if appropriate
[pydantic](https://pydantic-docs.helpmanual.io/) using native python 3 type hints to apply data validation at runtime. This could make validating loaded configs much easier. It also provides the ability to export JSON schema that can be exported and used in editor plugins and linting stages of pipelines.

This would also replace schematics (inherited by stacker) with a modern take on the same functionality.

ITProKyle:
## Why This Is Needed

resolves #431 

### Added

- added data classes for Runway & CFNgin configurations including dedicated data classes for each complex component
	- provide type checking when parsing the configuration files and through processing as values are resolved
	- transforms data as needed
	- can be used to export a JSON schema
	- raises a detailed error pinpointing invalid configuration

### Changed

- CFNgin package cache directory is now `./.runway/cache`
- the runway config class is now `runway.config.RunwayConfig`
- renamed and moved Runway config component classes
- Runway config component (definition) classes no longer handle data parsing and instead use a data class
	- they now act as a higher level API for working with the config
- the CFNgin config class is now `runway.config.CfnginConfig`
	- renamed and moved config component classes
- updated documentation for Runway and CFNgin configuration

### Removed

- removed support for `stacker*` keys in the CFNgin config
- removed support for legacy environment mappings in fields such as `env_vars`, `parameters`, etc
- removed support for legacy `environments` usage (functionally that was moved to `parameters`)


ITProKyle:
I am going to merge this in for now to reduce the amount of rebase catchup required to keep it up to date.

After having used used this branch and the schema it produces for a while, I think it is at least vary close to feature parity if not already there. (I fixed some of the annoyances I had noticed)

If any future changes are needed, they can be addressed individually.

Issue #310, opened 2020-05-19
[TODO] implement the use of pathlib for working with files/dirs


ITProKyle:
This should only be done once we drop python 2 (#295) for full support of `Path` objects when used with built-ins

- replaces the need for many `os.path` operations like checking if it is a file/dir, path manipulation, and listing/walking a dir
- replaces uses of open since a `Path` object and be read from/written to directly
- overall, `Path` objects are easier to work with

ITProKyle:
this has been completed in _most_ places

PR #463, opened 2020-09-24
staticsite: drop some posix path assumptions

troyready:
This almost certainly isn't causing any issues for anyone now, but it's nice to have consistency.


ITProKyle:
@troyready - FYI cases like this are a big reason to move away from using `os.path` for path/file manipulation and instead use `pathlib.Path` which removes the need to worry about OS specific path differences like this.

Less of something to change here and now since there are some extra steps when using Path objects to support python 2 but an initiative for v2 to remove all `os.path` usage (#310).

PR #464, opened 2020-09-25
fix non-idempotent Auth@Edge function deploys

troyready:
pyrsa/pyjwt create scripts with a shebang specific to the build system. Removing them, and setuptools remnants containing various hash values, stops new functions from being built on different systems.

It's possible the `PYTHONDONTWRITEBYTECODE` option is safe to apply unconditionally to all function builds by default, but I'm not 100% certain of the implications of that so I've gated it to an explicit option for now.

All of the new options have been intentionally left off the docs for this first pass where we may revise the names.

Discrepancies in the packages looked like:
```
diff -ur old/ecdsa-0.15.dist-info/RECORD new/ecdsa-0.15.dist-info/RECORD
--- old/ecdsa-0.15.dist-info/RECORD	2020-09-24 18:52:58.000000000 -0700
+++ new/ecdsa-0.15.dist-info/RECORD	2020-09-24 11:38:08.000000000 -0700
@@ -2,33 +2,32 @@
 ecdsa-0.15.dist-info/LICENSE,sha256=PsqYRXc9LluMydjBGdNF8ApIBuS9Zg1KPWzfnA6di7I,1147
 ecdsa-0.15.dist-info/METADATA,sha256=Vipd5pI4sqqaWMjmDzRNRkZCQaq1YDHOHkAJPlI92tw,24899
 ecdsa-0.15.dist-info/RECORD,,
-ecdsa-0.15.dist-info/REQUESTED,sha256=47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU,0
 ecdsa-0.15.dist-info/WHEEL,sha256=8zNYZbwQSXoB9IfXOjPfeNwvAsALAjffgk27FqvCWbo,110
 ecdsa-0.15.dist-info/top_level.txt,sha256=7ovPHfAPyTou19f8gOSbHm6B9dGjTibWolcCB7Zjovs,6
 ecdsa/__init__.py,sha256=3wbqSX9mkjn_sjkbx2vU-MJbKg0uz8DYLAZE5Jk4iyc,1219
-ecdsa/__pycache__/__init__.cpython-37.pyc,,
-ecdsa/__pycache__/_compat.cpython-37.pyc,,
-ecdsa/__pycache__/_rwlock.cpython-37.pyc,,
-ecdsa/__pycache__/_version.cpython-37.pyc,,
-ecdsa/__pycache__/curves.cpython-37.pyc,,
-ecdsa/__pycache__/der.cpython-37.pyc,,
-ecdsa/__pycache__/ecdh.cpython-37.pyc,,
-ecdsa/__pycache__/ecdsa.cpython-37.pyc,,
-ecdsa/__pycache__/ellipticcurve.cpython-37.pyc,,
-ecdsa/__pycache__/keys.cpython-37.pyc,,
-ecdsa/__pycache__/numbertheory.cpython-37.pyc,,
-ecdsa/__pycache__/rfc6979.cpython-37.pyc,,
-ecdsa/__pycache__/test_der.cpython-37.pyc,,
-ecdsa/__pycache__/test_ecdh.cpython-37.pyc,,
-ecdsa/__pycache__/test_ecdsa.cpython-37.pyc,,
-ecdsa/__pycache__/test_ellipticcurve.cpython-37.pyc,,
-ecdsa/__pycache__/test_jacobi.cpython-37.pyc,,
-ecdsa/__pycache__/test_keys.cpython-37.pyc,,
-ecdsa/__pycache__/test_malformed_sigs.cpython-37.pyc,,
-ecdsa/__pycache__/test_numbertheory.cpython-37.pyc,,
-ecdsa/__pycache__/test_pyecdsa.cpython-37.pyc,,
-ecdsa/__pycache__/test_rw_lock.cpython-37.pyc,,
-ecdsa/__pycache__/util.cpython-37.pyc,,
+ecdsa/__pycache__/__init__.cpython-38.pyc,,
+ecdsa/__pycache__/_compat.cpython-38.pyc,,
+ecdsa/__pycache__/_rwlock.cpython-38.pyc,,
+ecdsa/__pycache__/_version.cpython-38.pyc,,
+ecdsa/__pycache__/curves.cpython-38.pyc,,
```

&
```
diff -ur old/bin/pyrsa-verify new/bin/pyrsa-verify
--- old/bin/pyrsa-verify	2020-09-24 18:52:58.000000000 -0700
+++ new/bin/pyrsa-verify	2020-09-24 11:38:08.000000000 -0700
@@ -1,4 +1,4 @@
-#!/root/.local/share/virtualenvs/build-3vGKWv3F/bin/python
+#!/home/troyready/onica/bitbucket/myapp/.venv/bin/python
 # -*- coding: utf-8 -*-
 import re
 import sys
```

### DIFF
--- a/runway/cfngin/hooks/aws_lambda.py
+++ b/runway/cfngin/hooks/aws_lambda.py
@@ -3,6 +3,7 @@ import hashlib
 import json
 import logging
 import os
+import shutil
 import stat
 import subprocess
 import sys
@@ -455,12 +456,17 @@ def dockerized_pip(
 
     LOGGER.info('using docker image "%s" to build deployment package...', docker_image)
 
+    docker_run_args = {}
+    if _kwargs.get("python_dontwritebytecode"):
+        docker_run_args["environment"] = "1"
+
     container = client.containers.run(
         image=docker_image,
         command=["/bin/sh", "-c", pip_cmd],
         auto_remove=True,
         detach=True,
         mounts=[work_dir_mount],
+        **docker_run_args
     )
 
     # 'stream' creates a blocking generator that allows for real-time logs.
@@ -582,6 +588,10 @@ def _zip_package(
                 "--no-color",
             ]
 
+            subprocess_args = {}
+            if kwargs.get("python_dontwritebytecode"):
+                subprocess_args["env"] = dict(os.environ, PYTHONDONTWRITEBYTECODE="1")
+
             # Pyinstaller build or explicit python path
             if getattr(sys, "frozen", False) and not python_path:
                 script_contents = os.linesep.join(
@@ -612,12 +622,23 @@ def _zip_package(
             )
 
             try:
-                subprocess.check_call(cmd)
+                subprocess.check_call(cmd, **subprocess_args)
             except subprocess.CalledProcessError:
                 raise PipError
             finally:
                 if tmp_script.is_file():
                     tmp_script.unlink()
+
+        if kwargs.get("python_exclude_bin_dir") and os.path.isdir(
+            os.path.join(tmpdir, "bin")
+        ):
+            LOGGER.debug("Removing python /bin directory from Lambda files")
+            shutil.rmtree(os.path.join(tmpdir, "bin"))
+        if kwargs.get("python_exclude_setuptools_dirs"):
+            for i in os.listdir(tmpdir):
+                if i.endswith(".egg-info") or i.endswith(".dist-info"):
+                    LOGGER.debug("Removing directory %s from Lambda files", i)
+                    shutil.rmtree(os.path.join(tmpdir, i))
 
         req_files = _find_files(tmpdir, includes="**", follow_symlinks=False)
         return _zip_files(req_files, tmpdir)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Static Site Auth@Edge: fix non-idempotent deploys when deploying from different systems
 
 ## [1.14.0] - 2020-09-23
 ### Changed
--- a/runway/hooks/staticsite/auth_at_edge/lambda_config.py
+++ b/runway/hooks/staticsite/auth_at_edge/lambda_config.py
@@ -112,7 +112,14 @@ def write(
                 context,
                 provider,
                 bucket=kwargs["bucket"],
-                functions={handler: {"path": dirpath}},
+                functions={
+                    handler: {
+                        "path": dirpath,
+                        "python_dontwritebytecode": True,
+                        "python_exclude_bin_dir": True,
+                        "python_exclude_setuptools_dirs": True,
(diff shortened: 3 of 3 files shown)


troyready:
New zip:
image

Old zip with PYTHONDONTWRITEBYTECODE: 
image

Original zip:
image

troyready:
@ITProKyle feel free to make any changes directly or assign a revising issue to me for anything you'd prefer to be different.

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -455,12 +456,17 @@ def dockerized_pip(
 
     LOGGER.info('using docker image "%s" to build deployment package...', docker_image)
 
+    docker_run_args = {}
+    if _kwargs.get("python_dontwritebytecode"):
```
ITProKyle, line 460: since kwargs is being used now, the proceeding underscore should be removed as it indicates an unused variable.

runway/cfngin/hooks/aws_lambda.py
```diff
@@ -582,6 +588,10 @@ def _zip_package(
                 "--no-color",
             ]
 
+            subprocess_args = {}
+            if kwargs.get("python_dontwritebytecode"):
```
ITProKyle, line 592: we could instead expose this as an `env_vars`/`environ` argument for more flexibility. I don't think it would hurt to have something like this set throughout all of the hook's actions.

ITProKyle (commented):
For a first pass/patch this should be fine. Only a few minor changes i would make.

As for the deletion of `bin` and `*-info`, I would rather see it mimic `serverless-python-requirements` (partially to simplify moving between the two tools) like we did for the docker implementation. They use [slim](https://github.com/UnitedIncome/serverless-python-requirements#slim-package) (default set of globs to remove) with optional extra glob patterns to give a bit more control.

```yaml
slim: true
slimPatternsAppendDefaults: true
slimPatterns:
  - '**/*.egg-info*'
```

I put this all here so I could comment directly on the code. I can create issues later that link back to the comments.

runway/cfngin/hooks/aws_lambda.py
troyready: Legit embarrassing that I copied and pasted and didn't even think about that.

runway/cfngin/hooks/aws_lambda.py
troyready: You're probably right. I'm pretty torn on it between an implementation detail to hide away vs a configuration option.

Would be really nice if there was reasonable assurance that it wouldn't break things to just bake it in by default; I'm not certain what the implications with different types of packages/platforms.

ITProKyle: > just bake it in by default; I'm not certain what the implications

I'm not sure either. But, I can guarantee you someone will fine some package that will be incompatible.

We could still implement it as an environment variables option but, pre-populate it with values (like this one) by default and allow users to override the default by specifying their own to provide an out if it does break things.

ITProKyle: This was added to get rid of the `.pyc` files correct? The `slim` option I mentioned and linked above looks to do this by just deleting the files rather than using the env var - `.**/*.py[c|o]`, `**/__pycache__*`, and `**/*.dist-info*` are the default globs.

PR #493, opened 2020-12-09
add docker cfngin hook

### LINKED ISSUE #490: [REQUEST] ECR image hook
With the release of public ECR (in response to Docker Hub limits) and the ability to use ECR images as Lambda function deployment package ECR is now a very appealing option. There is a void in capabilities to build, test, and push images.

## Acceptance Criteria

- [x] build docker image
- [x] ~optionally test the image~ leaving this out for now, until there is a need for it
- [x] push the image that was built
- [x] push image that has already been built
- [ ] delete image locally and/or from ECR
- [x] provide access to sane docker options in the build/push process
- [ ] return information relevant to use as Lambda function deployment package and with other services

ITProKyle:
## Summary

Adds a hook to `docker image build` and `docker image push` with handling to simplify usage with AWS Elastic Container Registry.

## Why This Is Needed

resolves #490 

### Added

- add `runway.core.providers.docker` to extend the functionality of `docker-py`
- add `runway.lookups.handlers.ecr` lookup to get information from either the runway or CFNgin config
	- currently only supports a query of `login-password` which returns the same value as the awscli command `aws ecr get-login-password`
- add `runway.cfngin.hooks.docker` to interact with docker by mimicking the functionality of the docker CLI. The following actions are currently supported
	- `docker.login`
		- options to simplify ECR authentication
	- `docker.image.build`
		- options to simplify building an image of ECR
	- `docker.image.push`
		- options to simplify pushing an image to ECR
- add `runway.cfngin.hooks.ecr.purge_repository` to remove all images in an ECR repository so it can be deleted by CloudFormation


runway/core/providers/docker/models/images.py
```diff
@@ -0,0 +1,129 @@
...
+        result_stream, internal_stream = itertools.tee(json_stream(resp))
+        # added real-time logging
+        for chunk in internal_stream:
+            if "error" in chunk:
+                LOGGER.error(chunk["error"].strip())
+                raise BuildError(chunk["error"], result_stream)
+            if "stream" in chunk:
+                stream_msg = chunk["stream"]
+                if re.search(r"^Step \d*\/\d* : \w", stream_msg):
+                    LOGGER.info(stream_msg.strip())
```
troyready, line 114: Can we PR in these logging options upstream, instead of vendoring it here? I get how much more informative it is, but it'd  be really nice to just be able to call the library methods directly.

ITProKyle: I agree that it would be nice but I highly doubt it would be accepted upstream. They don't log anything outside of internal debug messages (nothing from the Docker service itself). _Most_ of the SDK returns a generator to iterate over for real-time logs. This one does not since it needs to inspect the logs to get the image ID.

ITProKyle: > They don't log anything outside of internal debug messages

This was determined by searching their codebase for uses of `logging.getLogger` (used in 6 places). They assign their logger to the variable `log`. Searching for `log.error`, `log.warning`, and `log.error` returned no results. Searching for `log.debug` returned 5 results, none of which log anything returned from the Docker API. Only internal messages.

@troyready - if you want me to try submitting a PR upstream I can but given the above it probably won't go anywhere. Just trying to prevent this from being a blocker here or in the docker-py repo. I can remove the logging completely if it's an issue. It's just not as "user friendly" especially for long running builds with a lot of steps (or those of us that panic a little when the terminal stops providing feedback for more than 3 seconds) as the terminal will just sit there and not be able to provide feedback until the `docker image build` action is complete.

troyready: Sorry for the delayed response.

Please do remove the logging. I don't think there's a perfect answer here, but I'd rather future us have a bit more distance from the internals of docker-py.

troyready (approved):

PR #511, opened 2021-01-19
relocate context classes & update them to share the same base class

### LINKED ISSUE #302: [REQUEST] both types of context objects should share a common base class
# Acceptance Criteria

- [ ] runway and cfngin contexts should share a common base class
- [ ] consolidate methods, attributes, args, etc as much as possible to create a unified experience when using either context object
	- this will require renaming methods/attributes that can and will break cfngin hooks, lookups, and blueprints
- [ ] 100% test coverage of the context objects

ITProKyle:
## Summary

Update the Runway and CFNgin context classes so that they share the same base class which contains shared methods, attributes, and properties.

Rename both classes so that it's easier to distinguish which context is being worked with.

Relocate both context objects to be importable from a similar path.

## Why This Is Needed

resolves #302 

### Added

- added `deploy_environment` arg and `.env` attribute to `CfnginContext`

### Changed

- relocate runway and cfngin context to one location
- renamed both context types to reflect what they belong to
- changed `CfnginContext.environment` to `CfnginContext.parameters` so that the attribute more accurately describes what it contains
- update tests
- update docs where needed

### Removed

- removed `CfnginContext.region` as it can now be access via the new `.env` attribute (same as `RunwayContext`)
- removed deprecated `RunwayContext.env_*` properties that aliased into `.env`
- removed `boto3_credentials` arg and its protected attribute from `CfnginContext`


ITProKyle:
merging this for now to keep things moving. anything that needs to be modified can be addressed in a future PR.

PR #524, opened 2021-01-21
improve module base classes, move module base classes and utilities

### LINKED ISSUE #311: [REQUEST] update `runway.module` base class with more attributes & methods
As was started in #268 with `RunwayModuleNpm`, the `RunwayModule` base class should unpack the `options` arg or even accept its items as `**kwargs` to make interaction with each component easier. We should also identify any commonly used functions and add them as `methods`.

This will break any custom modules.

# Acceptance Criteria

- [x] `RunwayModule` should have the following properties
	- [x] ~environment~ explicitly_enabled
	- [x] options (only contains `module.options`; assume subclasses with be parsing the options and assigning their own `options` attribute so maybe store as `self._options` or similar with the base class)
	- [x] parameters
	- [x] path as a `Path` object (#310 + #295)
- [x] update all module classes to account for data being rearranged and to support the change to a `Path` object
- [x] identify frequently used functions and add them as methods of the base class (e.g. `check_for_npm` & `npm_install` being added to `RunwayModuleNpm`)
- [x] 100% test coverage

ITProKyle:
## Summary

The `runway.module.base.RunwayModule` replaces `runway.module.RunwayModule` to provide more useful methods and attributes where previously all information was being stored in the `options` attribute of the class.

This also relocates all classes, functions, etc from `runway.module.__init__` to `runway.module.base` and `runway.module.utils` to avoid any future issues that can arise from the pattern.

## Why This Is Needed

resolves #311 

### Changed

- moved all classes, functions, etc from `runway.module.__init__` to `runway.module.base` and `runway.module.utils`
- `RunwayModule.options` now only contains `options`. `parameters`, etc each have their own attribute
- updated all module type classes to use `runway.module.base.RunwayModule` as their base class
- updated `runway.core.components.Module` to pass args appropriately to subclasses of `runway.module.base.RunwayModule`
- consolidate some utility functions in `RunwayModuleNpm` methods/static methods

### Removed

- removed functions that were made redundant


ITProKyle:
merging this for now. if anything needs to be changed, it can be completed in a followup PR.

PR #544, opened 2021-02-24
fix lambda zip errors when docker creates files as root

troyready:
aws_lambda hook can generate an error like:

```
[runway] method runway.cfngin.hooks.aws_lambda.upload_lambda_functions threw an exception
Traceback (most recent call last):
  File "/usr/lib/python3.8/shutil.py", line 670, in _rmtree_safe_fd
    os.unlink(entry.name, dir_fd=topfd)
PermissionError: [Errno 13] Permission denied: 'REQUESTED'
During handling of the above exception, another exception occurred:
...
PermissionError: [Errno 1] Operation not permitted: '/home/user/.runway_cache/cfnginy7x0_mye/tabulate-0.8.9.dist-info'
```

Runway shouldn't bomb out due to errors in cleaning up the temporary directory. This approach moves the cleanup to an explicit method that handles the edge case.

This looks in the diff like considerably more of a change than it is - every line isn't really modified, it's just:

* moving the context object to a function-scoped variable of the same name
* updating references to the object to call its `name` property
* adding an explicit cleanup at the end

### DIFF
--- a/runway/cfngin/hooks/aws_lambda.py
+++ b/runway/cfngin/hooks/aws_lambda.py
@@ -516,7 +516,8 @@ def _pip_has_no_color_option(python_path):
     return False
 
 
-def _zip_package(
+# too-many-statements override can be dropped with py2 support
+def _zip_package(  # pylint: disable=too-many-statements
     package_root,
     includes,
     excludes=None,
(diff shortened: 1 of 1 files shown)


troyready:
Merging to get a fix out now. Open to different approaches (e.g. going back to the context object wrapped in `try`, better messaging) in the future

Issue #546, opened 2021-02-26
HCL parser errors aren't caught and don't give useful debugging information


troyready:
There are 2 primary cases where both parsers will bomb out:

1) Deploying a Terraform module with a bad config
2) Deploying a module with some new HCL syntax not supported by pyhcl2

I think we can improve the experience for both cases. For bad configs, it's useful to print the exception to give info on the error. For syntax errors (hopefully less common in the future), the exception is useful in pointing to syntax that can be adjusted to bypass the error.

troyready:
Stacktrace from a recent deployment with some legitimately bad Terraform config:

```
VERBOSE:runway.env_mgr.tfenv:failed to parse as HCL2; trying HCL
Traceback (most recent call last):
  File "/home/troyready/.local/share/virtualenvs/k8s-tf-infrastructure-IjwjZZQD/lib/python3.9/site-packages/lark/parsers/lalr_parser.py", line 59, in get_action
    return states[state][token.type]
KeyError: 'LESSTHAN'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/troyready/.local/share/virtualenvs/k8s-tf-infrastructure-IjwjZZQD/lib/python3.9/site-packages/runway/env_mgr/tfenv.py", line 191, in terraform_block
    result = load_terrafrom_module(hcl2, self.path).get("terraform", {})
  File "/home/troyready/.local/share/virtualenvs/k8s-tf-infrastructure-IjwjZZQD/lib/python3.9/site-packages/runway/env_mgr/tfenv.py", line 122, in load_terrafrom_module
    tf_config = parser.loads(tf_file.read_text())
  File "/home/troyready/.local/share/virtualenvs/k8s-tf-infrastructure-IjwjZZQD/lib/python3.9/site-packages/hcl2/api.py", line 18, in loads
    return hcl2.parse(text + "\n")
  File "/home/troyready/.local/share/virtualenvs/k8s-tf-infrastructure-IjwjZZQD/lib/python3.9/site-packages/lark/lark.py", line 464, in parse
    return self.parser.parse(text, start=start)
  File "/home/troyready/.local/share/virtualenvs/k8s-tf-infrastructure-IjwjZZQD/lib/python3.9/site-packages/lark/parser_frontends.py", line 115, in parse
    return self._parse(token_stream, start)
[...]

PR #547, opened 2021-02-27
stop stacktrace on HCL parsing errors

troyready:
Highlights (hopefully) useful troubleshooting info.

Fixes: #546

(ignore the test file changes -- all that was find/replace on a cosmetic typo)

### DIFF
--- a/runway/env_mgr/tfenv.py
+++ b/runway/env_mgr/tfenv.py
@@ -103,7 +103,7 @@ def get_latest_tf_version(include_prerelease=False):
     return get_available_tf_versions(include_prerelease)[0]
 
 
-def load_terrafrom_module(parser, path):
+def load_terraform_module(parser, path):
     """Load all Terraform files in a module into one dict.
 
     Args:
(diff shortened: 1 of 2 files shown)


ITProKyle:
Can this wait until v2? The chunk of code is modified in the v2 branch so this will cause a ton of merge conflicts if it goes in now.

Instead of failing out, what are your thoughts on just continuing as is? We can still pretty print an error for the parsers as a warning but return an empty dict and another warning along the lines of `custom backend handling disabled; unable to parse files to determine backend`.

The need to parse the files and determine what backend is being used is only needed/was only added for handling remote backends which _should_ be a subset of use cases.

Going this route instead of what is in the PR, worst case scenario is that we don't handle a remote backend correctly and Terraform fails OR a user gets confused by the warning. As is in this PR, we are opening ourselves up to needed to support all sorts of HCL/HCL2 parsing errors that we have no control over.

ITProKyle:
superseded by #563

Issue #548, opened 2021-02-27
kbenv executions can fail with system aws-cli v2 installed


troyready:
`kbenv` commands, when invoking `aws` for authentication, will fail if the launched aws-cli is v2 with a `'Namespace' object has no attribute 'cli_binary_format'` error.

To reproduce (tested on macOS Catalina):
* Install runway and aws-cli v2 (ensure v2 is first in the path before any other aws-cli installs)
* Deploy the k8s repo generated via `runway gen-sample k8s-tf-repo`
* Attempt to run the `kbenv` commands [in the example](https://github.com/onicagroup/runway/tree/v1.18.2/runway/templates/k8s-tf-repo#hello-world-app)

I verified that setting [cli_binary_format config option](https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-cli_binary_format.html) doesn't avoid the error

PR #549, opened 2021-02-27
avoid awscli instantiation during kbenv

troyready:
This feels a tiny bit gross, but shouldn't have any performance issues at any point we'd care about and doesn't break any functionality that I can find.

Avoids weird interactions with runway awscli v1 and system installs of awscli v2.

Fixes: #548


troyready:
I'm going to merge this now to get the fix out. Open to reverting this and fixing #548 differently in the future, but don't want to leave it broken while other options are considered.

PR #563, opened 2021-03-25
tfenv: log hcl errors & return empty dict if module can't be parsed

### LINKED ISSUE #546: HCL parser errors aren't caught and don't give useful debugging information

ITProKyle:
## Summary

When an HCL parser fails, log the exception as a warning and include the path to the file that was being parsed at the time of the exception.

If both fail, return an empty dict which will cause Runway to treat the module as not having a backend configured in the scope of special backend handling. Right now, the only backend that requires special handling is [remote](https://www.terraform.io/docs/language/settings/backends/remote.html).

If special handling is required, Terraform will inevitably fail and the user can review the logs to see that Runway was unable to parse a file in the module to determine the backend so no special handling was applied.

## Why This Is Needed

Relying on third-party HCL/HCL2 parsers can result in flakey handling of Terraform modules where Terraform accepts the syntax in the files but our parsers do not.

resolves #546
supersedes #547 

### Added

- added `HclParserError`
- added warning when an HCL parser fails to parse a file; includes:
	- path to file that caused the error
	- reason for the error

### Changed

- Runway will no longer raise an error if both HCL and HCL2 parsers are unable to parse the files in the module; empty dict is used for Terraform block

## Screenshots

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,14 @@ help: ## show this message
 		printf "%s\n" $$help_info; \
 	done
 
+cov-report: ## display a report in the terminal of files missing coverage
+	@pipenv run coverage report \
+		--precision=2 \
+		--show-missing \
+		--skip-covered \
+		--skip-empty \
+		--rcfile=pyproject.toml
+
 list: ## list all targets in this Makefile
 	@$(MAKE) -pRrq -f $(lastword $(MAKEFILE_LIST)) : 2>/dev/null | awk -v RS= -F: '/^# File/,/^# Finished Make data base/ {if ($$1 !~ "^[#.]") {print $$1}}' | sort | egrep -v -e '^[^[:alnum:]]' -e '^$@$$'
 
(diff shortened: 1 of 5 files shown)


troyready (approved):
Seems great. The purpose/value of the overloaded generic functions isn't obvious to me, but I imagine it's relevant to the tests or something.

ITProKyle:
> Seems great. The purpose/value of the overloaded generic functions isn't obvious to me, but I imagine it's relevant to the tests or something.

It has no runtime impact - it's strictly for type checking.

troyready:
It's the duplicating the same types again at the end that just reads (from my outside novice perspective) as suboptimal to me; I'd almost rather leave Any spackled over them

ITProKyle:
> It's the duplicating the same types again at the end

Ya, I'm not a fan of it but the type checker complains if you don't 😞 . I'm sure there's probably some middle ground between strictness and not needing to add it but I have not found it yet.

Without adding type annotations to the real definition all the args are treated as `Unknown` throughout the function. If the overloaded types don't fit within the types on the real definition it complains that the overloads are invalid. You _could_ set the types on the real definition to `Any` but that can cause type errors in the logic - especially when using type narrowing to a generic as it will fill in the generics type variables with `Unknown` which fails type checking.

PR #571, opened 2021-04-02
remove `awscli` as a dependency

### LINKED ISSUE #550: kbenv interactions with awscliv2 need an integration test
#548 was resolved with a quick hotfix, but nothing is preventing the same issue from returning in the future.

ITProKyle:
## Summary

Having `awscli` as a dependency adds unnecessary complexity to Runway's dependencies when it's only being used to sync a directory to S3.

Before we can fully remove it, we need to replicate the functionality of `aws s3 sync`. This PR replicates the functionality with code derived from that of `awscli` to achieve the same level of performance.

resolves #550 

### Added

- added `Bucket.sync_from_local()` and `Bucket.sync_to_local()` for syncing files to/from an S3 bucket

### Removed

- removed `run-aws` command
- removed `awscli` as a dependency


ITProKyle:
updated `pyright` mid-PR so there are a few added changes for resolving type errors.

PR #595, opened 2021-05-05
fix sls tests

ITProKyle:
# Summary

Fix Serverless Framework tests and update dependencies.


ITProKyle:
Looks to be failing due to internal error - not the code itself.

image

PR #750, opened 2021-07-01
use flake8-print

ITProKyle:
# Summary

Use [flake8-print](https://github.com/JBKahn/flake8-print) to identify and remove unnecessary `print()` usage.


ITProKyle:
ignoring the decrease in coverage so no code functionality was changed - only some formatting/comments.

PR #1367, opened 2022-04-01
remove the upper `troposphere` requirement to allow `>=4`

ITProKyle:
# Why This Is Needed

To allow the use of `troposphere>=4.0.0` which adds type annotations.

# What Changed

## Changed

- changed troposphere requirement from `>=2.4, <4` to `>=2.4`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2375,7 +2375,7 @@ docs = ["dunamai", "jsx-lexer", "sphinx", "sphinx-github-changelog", "sphinx-rtd
 [metadata]
 lock-version = "1.1"
 python-versions = ">=3.7, <4"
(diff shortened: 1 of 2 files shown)


ITProKyle:
Functional tests are failing because secrets are not available to PRs from forks. The workflow should be updated so it does not trigger for PRs from forks.

danielxoi:
@mbordash2 Is it possible to do a small release to allow for this? There's a good bit of improvements on the `troposphere` side that would be great to leverage

Let me know if I can help somehow. I see some action work may be needed

mbordash2:
This has been addressed in a newer PR, closing.

Issue #2060, opened 2023-12-18
[BUG] I am trying to runway in my github actions. THe pipeline used to run well before but now its not working.


sagargulabani:
### Bug Description

I am running runway in my github actions pipeline. THis pipeline used to work well upto 2 months ago. However the pipeline has stopped working recently. When I dug deeper into the issue I found out the botocore wheel that is downloaded has services-2.json file as a compressed file, which means it is (services-2.json.gz). However as per the error, the code is looking for services-2.json

This looks like an issue with the newer versions of botocore after 1.32. 

Here is the entire error.

```
Run runway deploy --ci -e $DEPLOY_ENVIRONMENT --tag account:root
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.9.6/x64/bin/runway", line 5, in <module>
    from runway._cli.main import cli
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/runway/_cli/__init__.py", line 2, in <module>
    from .main import cli
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/runway/_cli/main.py", line 11, in <module>
    from . import commands, options
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/runway/_cli/commands/__init__.py", line 7, in <module>
    from ._gen_sample import gen_sample
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/runway/_cli/commands/_gen_sample/__init__.py", line [13](https://github.com/tech-speedyy/speedyy-management/actions/runs/7246506132/job/19738576590#step:7:14), in <module>
    from ._k8s_cfn_repo import k8s_cfn_repo
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/runway/_cli/commands/_gen_sample/_k8s_cfn_repo.py", line 12, in <module>
    from ....blueprints.k8s.k8s_workers import NodeGroup
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/runway/blueprints/k8s/k8s_workers.py", line 37, in <module>
    class NodeGroup(Blueprint):
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/runway/blueprints/k8s/k8s_workers.py", line 55, in NodeGroup
[...]

PR #2073, opened 2023-12-19
Adding logic to handle gzip'd botocore ec2 service-2.json

jake-skipper:
# Summary

Newer versions of `botocore` are now shipping with some files gzipped to save on storage space. Runway's K8s module depended on one of these files to determine available EC2 instance types, and could not read it as the path had changed (`service-2.json` vs `service-2.json.gz`).

# Why This Is Needed

Because of this upstream change, users who install a newer `botocore` version are unable to execute `runway` without an error.

# What Changed

Because we can't tell which version of `botocore` a user may have in their environment, logic has been added to check for which is available. In local testing this seems to work well. We were able to do a fresh `runway` install and confirm functionality with the gzipped file, then performed a `gunzip` on it and `runway` still operated fine with the uncompressed file.

# Checklist

- [x] Have you followed the guidelines in our [Contribution Requirements](https://docs.onica.com/projects/runway/page/developers/contributing.html)?
- [x] Does your submission pass tests?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your core changes, as applicable?
- [x] Have you updated documentation, as applicable?

### DIFF
--- a/runway/blueprints/k8s/k8s_workers.py
+++ b/runway/blueprints/k8s/k8s_workers.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 """Module with k8s nodegroup."""
+import gzip
 import json
 import os
 from typing import Any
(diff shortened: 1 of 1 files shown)


jake-skipper:
This PR is in regards to #2060

mbordash2 (approved):

PR #2381, opened 2024-06-21
update RTD config, move doc deps into their own group, update dev group

ITProKyle:
# What Changed

## Added

- **(actions)** added `info` job to the CI/CD workflow
  - uses github context to provide outputs that can be used to simplify conditionally skipping jobs
  - these outputs replace and improve some of the conditions used by the `test-functional` and `deploy-test-infrastructure` jobs as these are unable to run when a PR is opened from a fork because they require secrets

## Changed

- **(pyproject.toml)** renamed `[tool.poetry.dev-dependencies]` table to `[tool.poetry.group.dev.dependencies]`
  - new notation added in `poetry >= 1.2.0` [(source)](https://python-poetry.org/docs/managing-dependencies/#dependency-groups)
  - old notation will be deprecated in a future release
- **(pyproject.toml)** moved documentation dependencies out of the `main` dependency group into the `docs` dependency group (`[tool.poetry.group.docs.dependencies]` table)
  - updated `Makefile` accordingly to remove `--extras` flag
- **(pyproject.toml)** moved `boto3-stubs` & `mypy-boto3` out of the `dev` dependency group into the `types` dependency group to better reflect what they are being installed for
- **(.readthedocs.yml)** updated config to use the documented method to [Install dependencies with Poetry](https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry)
  - python version used to build documentation was also bumped when updating the config
- **(actions)** removed uses of `--extras docs`
- **(actions)** the `Spell Check` workflow will not trigger for pull requests and pushes to the `master` branch rather than only for pushes to all branches
  - this mirrors the triggering events of the `CI/CD ` workflow
  - could be the cause of the PR becoming stuck waiting for the `Spell Check` workflow to run/report status as this PR is opened from a fork which does not run actions by default


alexander-kogan (approved):
LGTM

ITProKyle:
codecov failure is a known bug in the current version of the action when creating a PR from a fork - https://github.com/codecov/codecov-action/issues/1487

mbordash2 (approved):

PR #2419, opened 2024-07-03
dependency updates for July

mbordash2:
# Summary
Dependency updates.

# Why This Is Needed
Keep project packages up to date.

# What Changed
Dependencies

## Changed
boto3 `1.34.129` -> `1.34.130`
requests `2.31.0` -> `2.32.3`
cspell `8.8.4` -> `8.10.0`
importlib-metadata `7.1.0` -> `8.0.0`
boto3-stubs `1.34.129` -> `1.34.139`
mypy-boto3 `1.34.129` -> `1.34.139`

# Checklist

- [x] Have you followed the guidelines in our [Contribution Requirements](https://docs.onica.com/projects/runway/page/developers/contributing.html)?
- [x] Does your submission pass tests?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your core changes, as applicable?
- [x] Have you updated documentation, as applicable?


package.json
```diff
@@ -53,13 +53,13 @@
   ],
   "version": "2.0.0-dev",
   "dependencies": {
-    "tar": "^7.2.0"
+    "tar": "^7.4.0"
```
ITProKyle, line 56: Does this need to be changed? All other changes in the PR are either lock file updates or dev dependency updates. Both of which would constitute adding the `changelog:skip` label as they have no impact on an end user. This changes the version requirement of a dependency which does impact the end user (although only for those installing from npm) which should be reflected in the title so it appears in release notes.

mbordash2: ok, I'll remove this from the PR.  We can do a separate one for the user impacting dependency updates.

ITProKyle (approved):

PR #2441, opened 2024-07-17
improve pre-commit hooks, replace flake8 & pylint with ruff

ITProKyle:
# Summary

Replace `flake8`, `isort`, and `pylint` with [`ruff`](https://docs.astral.sh/ruff/).

# What Changed

## Added

- **(pre-commit):** added `check-toml` hook
- **(pre-commit):** added `toml-sort` hook
- **(pre-commit):** added `mdformat` hook
- **(deps):** added [`ruff`](https://docs.astral.sh/ruff/)
  - added config
  - applied changes to satisfy `ruff`'s rules

## Changed

- **(pre-commit):** updated exclude/include for `pretty-format-json` to format more files
- **(pre-commit):** changed `files` to `exclude` for `yamlfmt` hook so that it handles automatic formatting for more files
- **(black):** `line-length` is now set to `100` to allow longer lines
- **(deps):** moved `black` to `lint` group

## Removed

- **(pre-commit):** removed `black` hook
  - while it can be nice to run `black` as a pre-commit hook, managing it's version and the versions of it's dependencies is difficult resulting in mismatches and errors
- **(pre-commit):** removed `isort` hook as `isort` is no longer in use
- **(pre-commit):** removed `flake8` hook as `flake8` is no longer in use
- **(deps):** removed use of `flake8`, `isort`, and `pylint`
  - removed all plugins
  - removed all configuration and comments related to these
- **(deps):** removed `backports.cached_property` as it will never be installed with current python version requirement
- **(deps):** removed `importlib-metadata` as it will never be installed with current python version requirement


mbordash2 (approved):
🚀

PR #2488, opened 2024-08-09
use `poetry-dynamic-versioning`, update workflows to consolidate steps

ITProKyle:
# Summary

Use the [`poetry-dynamic-versioning`](https://github.com/mtkennerly/poetry-dynamic-versioning) plugin to automatically set package version.
This plugin automates the use of [Dunamai](https://github.com/mtkennerly/dunamai) (what is currently used).

# What Changed

## Added

- added `runway.__version_tuple__` set by `poetry-dynamic-versioning`

## Changed

- `pipx` is now used instead of `pip` when installing `poetry` on the devcontainer image
  - `poetry-dynamic-versioning` is injected into the same venv as `poetry`
- standardized the names of workflow steps and added some emoji
- replaced the use of `actions/setup-python`, `Gr1N/setup-poetry`, and a few script steps with [`finleyfamily/action-setup-python`](https://github.com/finleyfamily/action-setup-python)
  - this is a composite action that combines setting up python, installing poetry, installing poetry plugins, installing dependencies (including caching), etc
- consolidated actions run for `pull_request_target` into a single workflow
- `runway.__version__` is no longer determined at runtime. it is set by `poetry-dynamic-versioning` when building a release
- default value of `runway_version` is now `None` instead of `>=1.10`
  - this is needed for tests to pass as `__version__` is statically set to `0.0.0` outside of the build & publish workflow 

## Removed

- removed `version` target from `Makefile` as it is no longer needed


alexander-kogan (approved):
LGTM

PR #2499, opened 2024-08-12
required version of `boto3` is now `^1.34.50`, remove dev dependencies from main group

ITProKyle:
# Summary

Removed dev dependencies from the `main` dependency group.
Upgrade required version of `boto3`

# Why This Is Needed

- dependencies that are not needed to run `runway` are being installed for users for `runway`
- the version requirement for `boto3` causes slow dependency resolution due to the large number of possible combinations of `boto3` and `botocore`
	- both packages a release every day or few days

# What Changed

## Changed

- upgraded required version of `boto3` from `^1.16` to `^1.34.50` (a number of patch releases behind the [latest version](https://pypi.org/project/boto3/1.34.158/))
- further split dependencies into different groups
- upgraded version requirements for dependencies not in the `main` group
- extended excepted branch names in the `Enforce Branch Name` job and updated the autolabel config to include these branch prefixes

## Removed

- removed `moto` from the `main` dependency group
- removed `pyinstaller` from the `main` dependency group
- removed `testfixtures` from the `main` dependency group
- removed `wheel` from the `main` dependency group
- removed `urllib3` from `main` dependency group as it is not a direct dependency

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -3,7 +3,7 @@ autolabeler:  # cspell:ignore autolabeler
   - label: bug
     branch: [/^(bug|bugfix|fix|hotfix)\/.*/]
   - label: changelog:skip
-    branch: [/^(dependabot)\/.*/]
+    branch: [/^(dependabot)\/.*/, /^pre-commit-ci-update-config/, /^renovate\/.*lockfile/]
   - label: cloudformation
     files:
       - '**/templates/*.json'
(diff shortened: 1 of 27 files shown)


ITProKyle:
Through process of elimination, updating the following set of dependencies in `poetry.lock` causes the `pipenv` tests to fail on Windows:

```
distlib
platformdirs
virtualenv
```

alexander-kogan (approved):
LGTM

Issue #2101, opened 2024-01-03
Upgrade Pydantic to v2


jake-skipper:
Runway uses Pydantic extensively, and awhile back they moved to v2. We haven't received those updates due to package pinning and some changes will be required to move to v2.

ITProKyle:
The decision was made to perform this upgrade, even though it is breaking, in a minor release rather than major.

Tasks to be completed:

- [ ] EoL of binary releases (pyinstaller) to S3 and npm
  - [ ] remove jobs that build, publish, etc these
- [ ] remove code and dependencies specific to pyinstaller 
- [ ] mark npm package as deprecated with message
- [ ] update install docs & README
- [ ] update pyright to the latest version now that it can be updated

PR #2508, opened 2024-08-15
end support for installing via cURL & npm

ITProKyle:
# Summary

End support for installing via cURL & npm. This enables us to upgrade to `pydantic` ^2 (#2101) without needed to produce `pyinstaller` builds per-architecture/platform.

# What Changed

## Changed

- updated `quickstarts/runway/Dockerfile`:
  - uses `ubuntu:24.04` as the base image
  - uses `pyenv` to install python
  - uses `pipx` to install runway with a not that recommends using poetry to install runway per-project instead

## Removed

- removed dependencies, documentation, files, and jobs related to `pyinstaller` builds
- removed documentation, files, and jobs related to publishing to S3 for installing runway via cURL
- removed documentation, files, and jobs related to publishing to npm


alexander-kogan (approved):
LGTM

PR #2547, opened 2024-09-05
delete files in `.terraform` rather than sending them to trash

### LINKED ISSUE #1952: [BUG] .Trash directory growing to large sizes with repeated terraform work
### Bug Description

On repeated executions involving Terraform deployments, Runway's repeated removal of Terraform files into `.Trash` can cause large amounts of disk usage.

### Expected Behavior

Running `plan` repeatedly shouldn't end up generating gigs of disk usage that isn't apparent to the end user. If `.Trash` functionality is needed, it should be configured in such a way as to not cause excessive use and be more apparent to end user what is happening.

### Steps To Reproduce

Repeatedly `runway plan` against a Terraform module and take note of the `.Trash` directory.

### Runway version

2.6.6

### Installation Type

pypi (pip, pipenv, poetry, etc)

### OS / Environment

N/A

### Anything else?

N/A

mbordash2:
# Summary
Rather than sending `.terraform` files to the `.Trash` directory, they are attempted to be deleted.  If this fails for some reason, the code falls back to the original functionality.  This doesn't impact the tool execution in any way, it just may clutter up the filesystem so failback should be fine.  

# Why This Is Needed

Users report that repeated runs of terraform will generate large amounts of files in the `.Trash` directory which may be undesirable and can cause file system full situations in pipelines, etc.

resolves #1952

# What Changed

## Changed

`.terraform` directory cleanup will remove files rather than move to trash unless deletion fails.

# Checklist

- [x] Have you followed the guidelines in our [Contribution Requirements](https://runway.readthedocs.io/page/developers/contributing.html)?
- [ ] Does your submission pass tests?
- [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your core changes, as applicable?
- [ ] Have you updated documentation, as applicable?

### DIFF
--- a/runway/module/terraform.py
+++ b/runway/module/terraform.py
@@ -12,7 +12,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
 import hcl
-from send2trash import send2trash
 from typing_extensions import Literal
 
 from .._logging import PrefixAdaptor
@@ -415,14 +414,10 @@ class Terraform(RunwayModule[TerraformOptions], DelCachedPropMixin):
                 self.logger.debug("directory retained: %s", child)
                 continue
             self.logger.debug("removing: %s", child)
-            if Path.exists(child):
-                try:
-                    Path.unlink(child)
-                except IsADirectoryError:
-                    shutil.rmtree(child)
-                except OSError:
-                    self.logger.warning("unable to remove %s, sending to .Trash", child)
-                    send2trash(child)  # does not support Path objects
+            if child.is_dir():
+                shutil.rmtree(child, ignore_errors=True)
+            else:
+                child.unlink(missing_ok=True)
 
     def deploy(self) -> None:
         """Run Terraform apply."""
--- a/runway/module/terraform.py
+++ b/runway/module/terraform.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import logging
 import re
+import shutil
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


runway/module/terraform.py
```diff
@@ -414,7 +415,14 @@ def cleanup_dot_terraform(self) -> None:
...
             self.logger.debug("removing: %s", child)
-            send2trash(str(child))  # does not support Path objects
+            if Path.exists(child):
+                try:
+                    Path.unlink(child)
+                except IsADirectoryError:
+                    shutil.rmtree(child)
+                except OSError:
+                    self.logger.warning("unable to remove %s, sending to .Trash", child)
+                    send2trash(child)  # does not support Path objects
```
ITProKyle, line 425:
```suggestion
            if child.is_dir():
                shutil.rmtree(child, ignore_errors=True)
            else:
                child.unlink(missing_ok=True)
```

- `child` is already a `Path` object so no need to pass it into `Path.exists` or `Path.unlink`
- check if its a directory first rather than catching an error raised if it is a directory
- the arguments passed in the suggestion ignore _most_ errors that would be raised except if its a file. in that case it only accounts for the file already being deleted
- removes the need for `send2trash` as a dependency (it can be removed with `poetry remove send2trash`)

mbordash2: sounds good, thanks!

ITProKyle (approved):

Issue #2405, opened 2024-06-28
[BUG] Fix exception pickling (cause of TypeError => BrokenProcessPool).


sam-fakhreddine:
Description: [BUG] Fix exception pickling (cause of TypeError => BrokenProcessPool).

mbordash2:
@ITProKyle to fix one as an example.

ITProKyle:
Heres an example of what needs to be done for all custom exceptions in the runway codebase:

1. All arguments must be positional.
2. If more than one argument is accepted, a `__reduce__` method must be added ([example `__reduce__` method plus a unit test for the exception](https://github.com/ITProKyle/runway/commit/ea843d7dfd06ba5a3bd518173f09f8e6e1491fa1)).

The base classes `RunwayError` and `CfnginError` are fine as written.

PR #2669, opened 2024-11-20
custom base exceptions updated for pickling

mbordash2:
# Summary

# Why This Is Needed
Implement exception pickling to prevent `TypeError`s as reported in issue #2405

# What Changed

## Fixed

Runway custom exceptions pickling implemented.

# Checklist

- [ ] Have you followed the guidelines in our [Contribution Requirements](https://runway.readthedocs.io/page/developers/contributing.html)?
- [ ] Have you checked to ensure there aren't other open [Pull Requests](../../../pulls) for the same update/change?
- [ ] Does your submission pass tests?
- [ ] Have you linted your code locally prior to submission?
- [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your core changes, as applicable?
- [ ] Have you successfully ran tests with your changes locally?

### DIFF
--- a/runway/exceptions.py
+++ b/runway/exceptions.py
@@ -37,9 +37,9 @@ class ConfigNotFound(RunwayError):
 
     looking_for: list[str]
     message: str
-    path: Path
+    path: Path | None
 
-    def __init__(self, *, looking_for: list[str] | None = None, path: Path) -> None:
+    def __init__(self, looking_for: list[str] | None = None, path: Path | None = None) -> None:
         """Instantiate class.
 
         Args:
@@ -56,6 +56,9 @@ class ConfigNotFound(RunwayError):
             self.message = f"config file not found at path {path}"
         super().__init__(self.path, self.looking_for)
 
+    def __reduce__(self) -> tuple[type[Exception], tuple[Any, ...]]:
+        """Support for pickling."""
+        return self.__class__, (self.looking_for, self.path)
 
 class DockerConnectionRefusedError(RunwayError):
     """Docker connection refused.
@@ -89,7 +92,9 @@ class DockerExecFailedError(RunwayError):
     exit_code: int
     """The ``StatusCode`` returned by Docker."""
 
-    def __init__(self, response: dict[str, Any]) -> None:
+    message: str = "error message undefined"
+
+    def __init__(self, response: dict[str, Any] | None = None) -> None:
         """Instantiate class.
 
         Args:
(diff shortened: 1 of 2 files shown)


runway/exceptions.py
```diff
@@ -99,9 +105,14 @@ def __init__(self, response: dict[str, Any]) -> None:
...
-        error: dict[Any, Any] = response.get("Error") or {}  # value from dict could be NoneType
-        self.message = error.get("Message", "error message undefined")
+        self.exit_code = (
+            response.get("StatusCode", 1) if response else 1
+        )  # we can assume this will be > 0
+        error: dict[Any, Any] | None = (
+            response.get("Error") if response else {}
+        )  # value from dict could be NoneType
+        if error:
+            self.message = error.get("Message", "error message undefined")
```
ITProKyle, line 115: This can be done without changing the signature by storing the `response` on the object.

```suggestion
        self._response = response
        self.exit_code = response.get("StatusCode", 1)  # we can assume this will be > 0
        error: dict[Any, Any] = response.get("Error") or {}  # value from dict could be NoneType
        self.message = error.get("Message", "error message undefined")
```

Then, the `__reduce__` method would return self.__class__, (self._response,)

ITProKyle (changes requested):

runway/exceptions.py
```diff
@@ -64,7 +68,7 @@ class DockerConnectionRefusedError(RunwayError):
 
     - Docker is not installed.
     - Docker service is not running.
-    - The current user does not have adequate permissions (e.g. not a member of
+    - The current user does not have adequate perons (e.g. not a member of
```
ITProKyle, line 71: ?

runway/cfngin/exceptions.py
```diff
@@ -250,7 +288,6 @@ def __init__(self, key: str, *args: Any, **kwargs: Any) -> None:
             **kwargs: Arbitrary keyword arguments.
 
         """
-        self.key = key
```
ITProKyle, line 253: this needs to be retained, it contributes to this - https://github.com/rackspace/runway/actions/runs/12190961499/job/34009033373#step:7:196

runway/cfngin/exceptions.py
```diff
@@ -602,21 +747,32 @@ def __init__(self, blueprint_name: str, *args: Any, **kwargs: Any) -> None:
             **kwargs: Arbitrary keyword arguments.
 
         """
+        self.args = args
+        self.kwargs = kwargs
```
ITProKyle, line 751: `args` & `kwargs` don't need to be saved since they arn't used (other than to pass to `super()`).

runway/cfngin/exceptions.py
ITProKyle: After reverting, add:

```python
    def __reduce__(self) -> tuple[type[Exception], tuple[Any, ...]]:
        """Support for pickling."""
        return self.__class__, (self.key,)
```

This allows the exception to reconstruct itself else the return of this method is essentially `self.__class__, (self.message,)` which is where the doubling in the string is coming from

ITProKyle (changes requested):

runway/exceptions.py
mbordash2: 😬

ITProKyle (changes requested):

ITProKyle (approved):

